PR #65, opened 2022-04-11
for loop gas efficiency 

cupOJoseph:
tiny gas savings with some hacks.
- [x] comments why we are doing it
- [x] passes hardhat test


songlh:
@cupOJoseph I have a question about this pr. Why incrementing the counter inside the loop can save gas?

PR #251, opened 2022-05-17
⚡️ Rewrite _verifyProof in assembly

transmissions11:


wilsoncusack:
wanna see that gas-snapshot diff!

PR #278, opened 2022-05-20
fix test function name

davidbrai:


0age (approved):
Good catch!

davidbrai:
@0age looks like this was closed but not merged. on purpose?

PR #282, opened 2022-05-21
Update Verifiers.sol

dannyadkins:
avoid returning valid, default to true

tests show 10-20 gas saved on both true and false conditions

### DIFF
--- a/contracts/lib/Verifiers.sol
+++ b/contracts/lib/Verifiers.sol
@@ -39,6 +39,9 @@ contract Verifiers is Assertions, SignatureVerification {
         uint256 endTime,
         bool revertOnInvalid
     ) internal view returns (bool valid) {
+        // Default to true if the order time is valid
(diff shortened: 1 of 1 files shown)


contracts/lib/Verifiers.sol
```diff
@@ -39,19 +39,20 @@ contract Verifiers is Assertions, SignatureVerification {
...
         if (startTime > block.timestamp || endTime <= block.timestamp) {
             // Only revert if revertOnInvalid has been supplied as true.
             if (revertOnInvalid) {
                 revert InvalidTime();
             }
 
-            // Return false as the order is invalid.
-            return false;
+            // Set to false as the order is invalid.
+            valid = false;
```
transmissions11, line 53: would returning false be any cheaper?

0age:
@dannyadkins can you reopen this PR so that CI will run?

PR #286, opened 2022-05-22
Ref seaport

cupOJoseph:
update reference to be named seaport instead of consideration

## Motivation

reference impl should be named `referenceSeaport` so everyone knows that it is the same as seaport

## Solution

rename ReferenceConsideration.sol to seaport and change everywhere in testing that it is used.


cupOJoseph:
hmmm linter failed but i ran prettier... let me see..

0age:
Closing as there are now a ton of merge conflicts, will be easier to reopen a new one!

PR #287, opened 2022-05-22
Vyper scaffolding

internnos:
## Motivation

Added support for Vyper


brownie-config.yaml
```diff
@@ -0,0 +1,66 @@
+project_structure:
+    build: build
```
0age, line 2: Could we put everything vyper-related into a vyper root folder?

brownie-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+compiler:
+    evm_version: null
+    solc:
+        version: null
+        optimizer:
+            enabled: true
+            runs: 200
+        remappings: null
+    vyper:
+        version: null
```
0age, line 38: Does this imply to use the latest version?

build/contracts/Greeter.json
```diff
@@ -0,0 +1,2302 @@
+{
```
0age, line 1: Let’s put the vyper build folder in .gitignore

build/tests.json
```diff
@@ -0,0 +1,15 @@
+{
```
0age, line 1: Same, gitignore’d please

0age (changes requested):
Looking good so far, left some comments

brownie-config.yaml
internnos: ofc, including the tests and contracts, or just the brownie-related files?

brownie-config.yaml
internnos: nope ser, it actually depends on the pragma used in the contract

This is taken from the [documentation](https://eth-brownie.readthedocs.io/en/stable/config.html#config-reports)

> The version of vyper to use. Should be given as a string in the format 0.x.x. If set to null, the version is set based on the contract pragma

build/contracts/Greeter.json
internnos: agree, on it

build/tests.json
internnos: on it

brownie-config.yaml
0age: Let’s do it all

brownie-config.yaml
0age: Perfect

brownie-config.yaml
internnos: like this?

```
project_structure:
    build: vyper/build
    contracts: vyper/contracts/
    interfaces: vyper/interfaces
    reports: vyper/reports
    scripts: vyper/scripts
    tests: vyper/test
```

0age: Exactly

brownie-config.yaml
```diff
@@ -0,0 +1,66 @@
+project_structure:
```
0age, line 1: Oh one other thing: can we put this file in the `config/` folder and specify it’s location via a flag somehow? Not a big deal if we can’t but want to keep the root level uncluttered where possible

brownie-config.yaml
internnos: done

brownie-config.yaml
internnos: oh yeah i was thinking about that too, but i haven't been able to do it until now

this is from the [documentation](https://eth-brownie.readthedocs.io/en/stable/config.html#config-reports)

> The configuration file must be saved as brownie-config.yaml. If saved in the root directory of a project it will be loaded whenever that project is active. If saved in your [home path](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home), it will always be loaded.


i'll keep trying, though

0age (approved):
LGTM!

banteg:
no need to copy over the entire brownie config, a usual config should be pretty minimal, something like:
```
networks:
    default: hardhat

project_structure:
    build: vyper/build
    contracts: vyper/contracts/
    interfaces: vyper/interfaces
    reports: vyper/reports
    scripts: vyper/scripts
    tests: vyper/test

autofetch_sources: true
```

if you already use hardhat, there is no need introduce another dependency of ganache, brownie supports hardhat properly.

for pytest+foundry, we'd likely need to migrate to ape later, but hold off for now, it doesn't support everything we need yet.

internnos:
> ### config
> no need to copy over the entire brownie config, a usual config should be pretty minimal, something like:
> 
> ```yaml
> networks:
>     default: hardhat
> 
> project_structure:
>     build: vyper/build
>     contracts: vyper/contracts/
>     interfaces: vyper/interfaces
>     reports: vyper/reports
>     scripts: vyper/scripts
>     tests: vyper/test
> 
> autofetch_sources: true
> ```
> 
> ### hardhat
> if you already use hardhat, there is no need introduce another dependency of ganache, brownie supports hardhat properly.
> 
> for pytest+foundry, we'd likely need to migrate to ape later, but hold off for now, it doesn't support everything we need yet.
> 
> ### isolation
> also add a vyper/test/conftest.py with this to get the test chain reverted to a state after all fixtures are set up after each test:
> 
> ```python
> import pytest
> 
> @pytest.fixture(autouse=True)
> def isolate(fn_isolation):
>     pass
> ```
> 
> read more [here](https://eth-brownie.readthedocs.io/en/stable/tests-pytest-intro.html#pytest-fixtures-isolation)
> 
> ### dev revert comments
> also note that brownie supports [dev revert strings](https://eth-brownie.readthedocs.io/en/stable/tests-pytest-intro.html#developer-revert-comments), e.g.
> 
> ```python
> assert water > 0, "no water in seaport"
> ```
> 
> can be written as a comment
> 
> ```python
> assert water > 0  # dev: no water in seaport
> ```
> 
> then it can be caught in tests like this:
> 
> ```python
> with brownie.reverts(dev_revert_msg="no water in seaport"):
>     seaport.check_water_levels()
> ```

gotcha, on it ser

banteg (approved):

storming0x (approved):
Changes LGTM, as part of this scope do you want to setup CI github actions?

this is one example we have, altho it may have too many steps not needed at this point.

https://github.com/yearn/yearn-vaults/blob/main/.github/workflows/test.yaml

0age:
> Changes LGTM, as part of this scope do you want to setup CI github actions?
> 
> this is one example we have, altho it may have some steps that arent needed at this point.
> 
> https://github.com/yearn/yearn-vaults/blob/main/.github/workflows/test.yaml

Let’s get this in on the next PR so that we unblock on setting up scaffolding!

PR #289, opened 2022-05-23
✨Gas golfing: make constructors `payable`

pcaversaccio:
## Motivation

You can cut out 10 opcodes in the creation-time EVM bytecode if you declare a constructor `payable`. For more in-depth information see here: https://forum.openzeppelin.com/t/a-collection-of-gas-optimisation-tricks/19966/5.

## Solution

Make all constructors `payable` including the reference implementation due to the sake of consistency.


pcaversaccio:
Cc: @0age @transmissions11

PR #290, opened 2022-05-23
feat: rewrite EIP1271Wallet in vyper

banteg:


contracts/test/EIP1271Wallet.sol
```diff
@@ -1,111 +0,0 @@
-// SPDX-License-Identifier: MIT
```
0age, line 1: We definitely don't want to delete this, need to be able to run tests against solidity and also want it as a point of reference

contracts/test/EIP1271Wallet.vy
```diff
@@ -0,0 +1,78 @@
...
+
+
+@external
+def isValidSignature(digest: bytes32, signature: Bytes[65]) -> bytes4:
+    if self.digestApproved[digest]:
+        return _EIP_1271_MAGIC_VALUE
+    
+    assert len(signature) == 65
+
+    r: uint256 = convert(slice(signature, 0, 32), uint256)
```
0age, line 58: This syntax is _clean_

0age (changes requested):
Need to keep the solidity version, otherwise LGTM!

contracts/test/EIP1271Wallet.sol
banteg: multiple contracts with the same name would cause a name clash, need to think the project structure through

contracts/test/EIP1271Wallet.vy
```diff
@@ -0,0 +1,78 @@
...
+interface ERC20ApprovalInterface:
+    def approve(user: address, amount: uint256) -> bool: nonpayable
+
+interface NFTApprovalInterface:
+    def setApprovalForAll(user: address, approved: bool): nonpayable
+
+
+_EIP_1271_MAGIC_VALUE: constant(bytes4) = 0x1626ba7e
+
+owner: public(address)  # immutable
```
banteg, line 12: idk how to declare public immutable

internnos (commented):
super crisp, however i thought that this should be put inside `vyper/contracts` instead of contracts? or am i missing something?

contracts/test/EIP1271Wallet.vy
fubuloubu: Can't, but can make immutable constant and then reference in public view method

fubuloubu (approved):
Looks good

0age (approved):
LGTM

contracts/test/EIP1271Wallet.sol
0age: For all the reference contracts we prefixed them with `Reference`, could just prefix with `Vyper` for the time being

0age: Will it still cause a name clash as vyper build folder is isolated from the solidity cache?

0age:
> 

Oh that's absolutely right — going to move that now

PR #291, opened 2022-05-23
Update SeaportDocumentation.md

karmacoma-eth:
Fix typo as there are `ERC721_TO_ERC20` and `ERC1155_TO_ERC20` basic order types, but not the `_TO_ETH` variants

## Motivation

I believe there is a copy-paste error in the documentation

## Solution

Update the documentation to reflect the intended meaning


transmissions11 (approved):

transmissions11:
good find

PR #295, opened 2022-05-25
feat: rewrite AmountDeriver in Vyper

0xbok:


0age (approved):

vyper/contracts/libs/AmountDeriver.vy
```diff
@@ -0,0 +1,146 @@
...
+    # Multiply the numerator by the value and ensure no overflow occurs.
+    valueTimesNumerator: uint256 = val * numerator
+
+    # Divide that value by the denominator to get the new value.
+    newValue: uint256 = valueTimesNumerator / denominator
+
+    # Ensure that division gave a final result with no remainder.
+    exact: bool = ((newValue * denominator) / numerator) == val
+    if not exact:
+        raise "inexact fraction"
```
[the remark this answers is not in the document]
charles-cooper: or, inline the whole thing -- `assert ((newValue * denominator) / numerator) == val`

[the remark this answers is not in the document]
0xbok: inlined.

vyper/contracts/libs/AmountDeriver.vy
```diff
@@ -0,0 +1,144 @@
...
+            extraCeiling)
+
+        # Division is performed without zero check as it cannot be zero.
+        newAmount: uint256 = totalBeforeDivision / duration
+
+        # Return the current amount (expressed as endAmount internally).
+        return newAmount
+
+    # Return the original amount (now expressed as endAmount internally).
+    return endAmount
```
charles-cooper, line 63: nit: stylistically, might be good to fast path out `if startAmount == endAmount`. then the more complicated part can be less indented.

charles-cooper (approved):

vyper/contracts/libs/AmountDeriver.vy
0age: I also think `# Division is performed without zero check as it cannot be zero.` comment is maybe not true? Not sure whether vyper does a zero check on division, guessing it does though

charles-cooper: > I also think `# Division is performed without zero check as it cannot be zero.` comment is maybe not true? Not sure whether vyper does a zero check on division, guessing it does though

vyper division does do a 0 check. if you wanted to really skip the zero check, could use `unsafe_div(a, b)`

0xbok: I need to be more careful with the comments than just copying from the reference :) 
Will go with `unsafe_div`.

PR #296, opened 2022-05-25
Vyper hardhat

internnos:
## Motivation

Run current tests against Vyper contracts using `TARGET='vyper'` as environment variable


0age (changes requested):
Can we delete package-lock in favor of yarn? Otherwise this looks perfect

PR #297, opened 2022-05-25
feat: rewrite TokenTransferer in Vyper

de33:


vyper/contracts/TokenTransferer.vy
```diff
@@ -0,0 +1,75 @@
...
+        max_outsize=32
+    )  
+    if len(_response) > 0:
+        assert convert(_response, bool), "TransferFrom failed"  
+
+@internal
+def _performERC721Transfer(token: address, _from: address, to: address, identifier: uint256):
+    _response: Bytes[32] = raw_call(
+    token,
+    _abi_encode(
```
0xbok, line 33: discussed offline: we should go with using `IERC721(token).transfeFrom()` instead of `raw_call`.

vyper/contracts/TokenTransferer.vy
```diff
@@ -0,0 +1,75 @@
...
+        TRANSFERFROM1155_MID,
+        _from,
+        to,
+        identifier,
+        amount
+        ),
+        max_outsize=32
```
0age, line 55: same with 1155, no data

vyper/contracts/TokenTransferer.vy
```diff
@@ -0,0 +1,75 @@
...
+        _response: Bytes[32] = raw_call(
+        transfer.token,
+        _abi_encode(
+            SAFEBATCHTRANSFERFROM_MID,
+            transfer._from,
+            transfer.to,
+            transfer.ids,
+            transfer.amounts
+            ),
+            max_outsize=32
```
0age, line 72: same here, no data returned

vyper/contracts/TokenTransferer.vy
0age: I think so, that way codesize check will be performed and it'll be more readable

0age (changes requested):
left some comments (agree that casting to the token and performing the call that way is preferred)

vyper/contracts/TokenTransferer.vy
```diff
@@ -0,0 +1,32 @@
...
+struct ConduitBatch1155Transfer:
+    token: address
+    _from: address
+    to: address
+    ids: DynArray[uint256, 10] 
+    amounts: DynArray[uint256, 10] 
+
+@internal
+def _performERC20Transfer(token: address, _from: address, to: address, amount: uint256):
+    ERC20(token).transferFrom(_from, to, amount)
```
0age, line 19: This doesn’t check the return value though — need to revert on returning false

vyper/contracts/TokenTransferer.vy
```diff
@@ -0,0 +1,42 @@
...
+        token,
+        _abi_encode(
+            _from,
+            to,
+            amount,
+            method_id=method_id("transferFrom(address,address,uint256)")
+        ),
+        max_outsize=32
+    )  
+    if len(_response) > 0:
```
0age, line 28: If `len(_response) == 0` we also need a code size check

PR #298, opened 2022-05-26
Created a Vyper contract for `SignatureVerification`

Saw-mon-and-Natalie:
This PR converts `contracts/lib/SignatureVerification.sol` to `vyper/contracts/lib/SignatureVerification.vy`. 

As part of the new contract, 3 functions from `LowLevelHelpers.sol` were also included in `SignatureVerification.vy`. But only the `_staticcall` function was implemented due to not having low opcode access in Vyper.

- `_assertValidEIP1271Signature` and `_assertValidSignature` can have max `signature` length of 100 bytes; We can decide on a different upperbound.
- `_staticcall` can have a max `callData` length of 512 bytes; This upper bound can also be changed.

### DIFF
--- a/vyper/contracts/libs/SignatureVerification.vy
+++ b/vyper/contracts/libs/SignatureVerification.vy
@@ -8,55 +8,6 @@ IS_VALID_SIGNATURE_SELECTOR: constant(bytes4) = 0x1626ba7e
 EIP2098_allButHighestBitMask: constant(
     bytes32) = 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
 
-
-@internal
-@view
-def _staticcall(target: address, callData: Bytes[512]) -> (bool):
-    """
-     @dev Internal view function to staticcall an arbitrary target with given
-          calldata. Note that no data is written to memory and no contract
-          size check is performed.
-
-     @param target   The account to staticcall.
-     @param callData The calldata to supply when staticcalling the target.
-
-     @return success The status of the staticcall to the target.
-     """
-    success: bool = raw_call(
-        target,
-        callData,
-        is_static_call=True,
-        revert_on_failure=False
-    )
-    return success
-
-
-@internal
-@view
-def _revertWithReasonIfOneIsReturned():
-    """
-    @dev Internal view function to revert and pass along the revert reason if
-         data was returned by the last call and that the size of that data
-         does not exceed the currently allocated memory size.
-    """
-    raise "Not Implemented"
-
-
-@internal
-@pure
-def _doesNotMatchMagic(expected: bytes4) -> (bool):
-    """
-    @dev Internal pure function to determine if the first word of returndata
-        matches an expected magic value.
-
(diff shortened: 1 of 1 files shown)


vyper/contracts/libs/SignatureVerification.vy
```diff
@@ -0,0 +1,137 @@
...
+        digest: bytes32, signature: Bytes[100]) -> bytes4: view
+
+IS_VALID_SIGNATURE_SELECTOR: constant(bytes4) = 0x1626ba7e
+EIP2098_allButHighestBitMask: constant(
+    bytes32) = 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
+
+
+@internal
+@view
+def _staticcall(target: address, callData: Bytes[512]) -> (bool):
```
0age, line 14: Can delete this one as it's not actually being used by the Vyper implementation

vyper/contracts/libs/SignatureVerification.vy
```diff
@@ -0,0 +1,137 @@
...
+        callData,
+        is_static_call=True,
+        revert_on_failure=False
+    )
+    return success
+
+
+@internal
+@view
+def _revertWithReasonIfOneIsReturned():
```
0age, line 36: Same here

vyper/contracts/libs/SignatureVerification.vy
```diff
@@ -0,0 +1,137 @@
...
+    @dev Internal view function to revert and pass along the revert reason if
+         data was returned by the last call and that the size of that data
+         does not exceed the currently allocated memory size.
+    """
+    raise "Not Implemented"
+
+
+@internal
+@pure
+def _doesNotMatchMagic(expected: bytes4) -> (bool):
```
0age, line 47: Don't need this one either

vyper/contracts/libs/SignatureVerification.vy
```diff
@@ -0,0 +1,137 @@
...
+
+    @return A boolean indicating whether the expected value matches the one
+            located in the first word of returndata.
+    """
+    raise "Not Implemented"
+
+
+@internal
+@view
+def _assertValidEIP1271Signature(signer: address, digest: bytes32, signature: Bytes[100]):
```
0age, line 62: 100 bytes is actually quite limiting for the signature field; imagine cases where you want to supply a merkle proof as part of the signature

0age: Can always revisit this though

0age (changes requested):
The low-level helpers are actually unnecessary on this one as the 1271 check is done via a high-level call; otherwise LGTM!

vyper/contracts/libs/SignatureVerification.vy
Saw-mon-and-Natalie: I agree, that the upper bound needs to be revisited.

Saw-mon-and-Natalie:
@0age , removed the low-level internal functions.

PR #299, opened 2022-05-26
feat: rewrite ZoneInteraction in Vyper

tserg:
- Replaced enums with `uint256` (tracking support of enums: https://github.com/vyperlang/vyper/pull/2874)
- Omitted `_assertIsValidOrderStaticcallSuccess` function, which is replaced with inline assertions.

### DIFF
--- a/vyper/contracts/libs/ZoneInteraction.vy
+++ b/vyper/contracts/libs/ZoneInteraction.vy
@@ -1,0 +1,161 @@
+# @version ^0.3.3
+
+
+interface ZoneInteraction:
+    def isValidOrder(
+        orderHash: bytes32,
+        caller: address,
+        offerer: address,
+        zoneHash: bytes32
+    ) -> bytes4: view
+
+    def isValidOrderIncludingExtraData(
+        orderHash: bytes32,
+        caller: address,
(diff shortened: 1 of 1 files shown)


vyper/contracts/libs/ZoneInteraction.vy
```diff
@@ -0,0 +1,161 @@
...
+    signature: Bytes[1024]
+    extraData: Bytes[1024]
+
+
+struct CriteriaResolver:
+    orderIndex: uint256
+    side: uint256
+    index: uint256
+    identifier: uint256
+    criteriaProof: DynArray[bytes32, 10]
```
0age, line 65: Length 10 is insufficient for this one — say you have a 10k collection and you want to make an offer to buy any item that hasn't been compromised (where there are 10 compromised items): that means you have 9990 valid identifiers and a proof size of `Math.ceil(Math.log2(9990)) == 14`. I'd say a length of 15 would suffice for nearly all applications (though it's not ideal, length of 20 would be a more conservative choice)

0age (changes requested):
One comment, otherwise LGTM!

PR #302, opened 2022-05-27
New Vyper Contract for `ReenterancyGuard.vy`

Saw-mon-and-Natalie:
This PR implements the Vyper contract for `contracts/lib/ReenterancyGuard.sol`


banteg:
Vyper has a native reentrancy guard https://vyper.readthedocs.io/en/stable/control-structures.html#re-entrancy-locks

PR #311, opened 2022-06-04
Bump to 0.8.14, update conduit contracts, refactor test fixtures

d1ll0n:


contracts/lib/TokenTransferrerConstants.sol
```diff
@@ -33,6 +33,7 @@ pragma solidity >=0.8.7;
  *      codebase but have been left in for readability.
  */
 
+uint256 constant AlmostOneWord = 0x19;
```
leonardoalt, line 36: Huh shouldn't this be 31? 0x1f instead

gpersoon: how about `OneWord -1` ?

leonardoalt: better!

leonardoalt: Actually doesn't compile:
```solidity
TypeError: Only direct number constants and references to such constants are supported by inline assembly.
   --> contracts/lib/TokenTransferrer.sol:662:51:
    |
662 |                             add(returndatasize(), AlmostOneWord),
    |                                                   ^^^^^^^^^^^^^
```
So I kept it as 0x1f in the PR
https://github.com/ProjectOpenSea/seaport/pull/388

contracts/Seaport.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.13;
+pragma solidity >=0.8.13;
```
Thro77le, line 2: For caution and to avoid the unexpected, shouldn't the version be also upper bounded by `<0.9.0` or even `<=0.8.14`

PR #319, opened 2022-06-05
Fix partial fill overflow

d1ll0n:
Address partial fulfillment overflowing the stored numerator / denominator. Add additional tests for overflows and set IS_FIXED to true in the findings/PartialFillFractionOverflow test file.


0age (approved):

contracts/lib/ConsiderationConstants.sol
```diff
@@ -310,6 +310,8 @@ uint256 constant Conduit_execute_signature = (
     0x4ce34aa200000000000000000000000000000000000000000000000000000000
 );
 
+uint256 constant MaxUint120 = 0xffffffffffffffffffffffffffffff;
```
axic, line 313: This would be an option too:
```suggestion
uint256 constant MaxUint120 = uint256(type(uint120).max);
```

reference/lib/ReferenceOrderValidator.sol
```diff
@@ -238,7 +263,30 @@ contract ReferenceOrderValidator is
...
+     * @return greatestCommonDivisor The greatest common divisor.
+     */
+    function _greatestCommonDivisor(uint256 a, uint256 b)
+        internal
+        pure
+        returns (uint256 greatestCommonDivisor)
+    {
+        while (b > 0) {
+            uint256 c = b;
+            b = a % c;
```
axic, line 285: This will be checked against `c != 0`, but `c` can't be 0. Potentially use unchecked block for the mod.

contracts/lib/ConsiderationConstants.sol
0age: Nice readability improvement!

reference/lib/ReferenceOrderValidator.sol
0age: We've got a rule of no unchecked and no assembly in the reference implementation!

hrkrshnn:
The worst case for the GCD is `numerator` and `denominator` being 638817435613190341905763972389505493 and 1033628323428189498226463595560281832 respectively. Maybe useful to include it as a test?

contracts/lib/OrderValidator.sol
```diff
@@ -220,14 +222,64 @@ contract OrderValidator is Executor, ZoneInteraction {
...
+                            let _c := _b
+                            _b := mod(_a, _c)
+                            _a := _c
+                        }
+                        out := _a
+                    }
+                    let scaleDown := gcd(
+                        numerator,
+                        gcd(filledNumerator, denominator)
+                    )
```
gpersoon, line 251: Is it really necessary to do `gcd()` twice?  (costs gas and gets a possible worse value for `scaleDown`)
Because `numerator` isn't used in the rest of this function.
Except for `numerator := div(numerator, safeScaleDown)`, which isn't necessary then either.

contracts/lib/ConsiderationConstants.sol
naps62: For the record, this doesn't actually work.

```
error[7615]: TypeError: Only direct number constants and references to such constants are supported by inline assembly.
```

PR #334, opened 2022-06-06
Update CONTRIBUTORS.md

dabushori:
## Motivation

## Solution


dabushori:
Done! 👍

PR #336, opened 2022-06-06
Gas optimization, cache variable `orderParameters.offer.length`

eugenioclrc:
## Motivation
Gas optimization: caching the array length outside a loop saves reading it on each iteration, as long as the array's length is not changed during the loop.

## Solution
Cache the array length on a temp variable.

### DIFF
--- a/contracts/lib/OrderFulfiller.sol
+++ b/contracts/lib/OrderFulfiller.sol
@@ -211,8 +211,10 @@ contract OrderFulfiller is
                 }
             }
 
+            // Cache offers length
+            uint256 totalOffers = orderParameters.offer.length;
             // Iterate over each offer on the order.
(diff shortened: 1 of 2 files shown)


contracts/lib/OrderFulfiller.sol
```diff
@@ -211,8 +211,10 @@ contract OrderFulfiller is
                 }
             }
 
+            // Cache offers length
+            uint256 totalOffers = orderParameters.offer.length;
```
0age, line 215: Mind adding a newline after this one?

0age (changes requested):
One quick comment!

eugenioclrc:
Wow, its weird, i run the test before commit with the command `yarn test` and the all pass. Its seems that i forget to run the foundry tests.

PR #341, opened 2022-06-06
Enforce style prefix increment and postfix increment

eugenioclrc:
## Motivation

The difference between the prefix increment and postfix increment expression lies in the return value of the expression.

The prefix increment expression (`++i`) returns the updated value after it's incremented. The postfix increment expression (`i++`) returns the original value.

The prefix increment expression is cheaper in terms of gas.

Consider using the prefix increment expression whenever the return value is not needed.

Note to be careful using this optimization whenever the expression's return value is used afterwards, e.g. `uint a = i++` and `uint a = ++i` result in different values for `a`.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -6,3 +6,4 @@ Contributor                    | ENS
 d1ll0n                         | `d1ll0n.eth`
 transmissions11                | `t11s.eth`
 Kartik                         | `slokh.eth`
+0x4non                         | `punkdev.eth`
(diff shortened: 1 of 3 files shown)


0age:
mind running `yarn lint:fix` and commiting that?

eugenioclrc:
Ready!

0age:
Looks like another PR implemented this same thing ahead of yours, though yours has been open for a minute; want to close this one and just open another adding yourself as a contributor?

eugenioclrc:
@0age they frontrun me!!! we need a zero kwnoledge github LOL

PR #349, opened 2022-06-06
Change nested scope to unchecked nested scope

eugenioclrc:
## Motivation

All the calculations inside of the block where safe, so the calculations where wrapped in an unchecked block. There for its safe to move the unchecked scope to the parent scope.
Example;
```solidity
{
  for (uint i = 0; i < length;) {
     unchecked { /* math */ }
     // code
     unchecked { ++i; }
   }
}
```

This could be;
```solidity
unchecked {
  for (uint i = 0; i < length; ++i) {
     /* math */
     // code
   }
}
```

### DIFF
--- a/contracts/lib/OrderFulfiller.sol
+++ b/contracts/lib/OrderFulfiller.sol
@@ -283,7 +283,7 @@ contract OrderFulfiller is
          */
 
         // Declare a nested scope to minimize stack depth.
-        {
+        unchecked {
             // Declare virtual function pointer with ConsiderationItem argument.
             function(ConsiderationItem memory, address, bytes32, bytes memory)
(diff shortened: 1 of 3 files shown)


0age:
Can you run `yarn lint:fix` and commit again?

contracts/lib/OrderFulfiller.sol
```diff
@@ -193,7 +193,7 @@ contract OrderFulfiller is
          */
 
         // Declare a nested scope to minimize stack depth.
-        {
+        unchecked {
```
hrkrshnn, line 196: Somewhat arguable if the scope should be so wide. In the previous code, it was rather clear what was `unchecked`.

PR #350, opened 2022-06-06
minor optimizations

0xBeans:
## Motivation

An honest days work.

Small optimization of unchecked , changing to ++i or --i rather than += 1 or i -=1

### DIFF
--- a/reference/lib/ReferenceOrderCombiner.sol
+++ b/reference/lib/ReferenceOrderCombiner.sol
@@ -483,7 +483,7 @@ contract ReferenceOrderCombiner is
             if (execution.item.recipient == execution.offerer) {
                 unchecked {
                     // increment total filtered executions.
-                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256
+                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256.
                     ++totalFilteredExecutions;
                 }
             } else {
@@ -512,7 +512,7 @@ contract ReferenceOrderCombiner is
             if (execution.item.recipient == execution.offerer) {
                 unchecked {
                     // increment total filtered executions.
-                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256
+                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256.
                     ++totalFilteredExecutions;
                 }
             } else {
@@ -772,7 +772,7 @@ contract ReferenceOrderCombiner is
             if (execution.item.recipient == execution.offerer) {
                 unchecked {
                     // increment total filtered executions.
-                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256
+                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256.
                     ++totalFilteredExecutions;
                 }
             } else {
--- a/reference/lib/ReferenceOrderCombiner.sol
+++ b/reference/lib/ReferenceOrderCombiner.sol
@@ -238,9 +238,9 @@ contract ReferenceOrderCombiner is
             // Otherwise, track the order hash in question.
             orderHashes[i] = orderHash;
 
+            // Skip underflow check as maximumFulfilled is nonzero.
             unchecked {
                 // Decrement the number of fulfilled orders.
-                // maximumFulfilled == 0 check prevents underflow.
                 maximumFulfilled--;
             }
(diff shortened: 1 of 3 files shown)


reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -481,8 +481,10 @@ contract ReferenceOrderCombiner is
 
             // If offerer and recipient on the execution are the same...
             if (execution.item.recipient == execution.offerer) {
-                // increment total filtered executions.
-                totalFilteredExecutions += 1;
+                unchecked {
```
0age, line 484: Can you leave a comment as to why unchecked is safe here above the block and others? (i.e. that it is infeasible to increment that high)

0age (changes requested):

reference/lib/ReferenceOrderCombiner.sol
0xBeans: @0age Done.

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -481,8 +483,11 @@ contract ReferenceOrderCombiner is
...
+                    // increment total filtered executions.
+                    // totalFilteredExecutions starts at 0, infeasible to increment > 2^256.
```
0age, line 488: One tiny nit — let's ensure that comments do not cause the 80-character line limit to be exceeded

0age (changes requested):

reference/lib/ReferenceOrderCombiner.sol
0xBeans: Fixed :)

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -238,9 +238,11 @@ contract ReferenceOrderCombiner is
             // Otherwise, track the order hash in question.
             orderHashes[i] = orderHash;
 
-            // Decrement the number of fulfilled orders.
-            maximumFulfilled--;
-
+            unchecked {
```
0age, line 241: Tiny nit: move the comment as to why unchecked is safe to right before the unchecked block and revise to `// Skip overflow check as maximumFulfilled is nonzero.`

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -507,8 +512,11 @@ contract ReferenceOrderCombiner is
 
             // If offerer and recipient on the execution are the same...
             if (execution.item.recipient == execution.offerer) {
-                // increment total filtered executions.
-                totalFilteredExecutions += 1;
+                unchecked {
+                    // increment total filtered executions.
+                    // executions start at 0, infeasible to increment > 2^256.
```
0age, line 517: Nit: Capitalize Executions and place that word at the end of the comment line above.

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -764,8 +772,11 @@ contract ReferenceOrderCombiner is
...
```
0age, line 777: Nit: Capitalize Executions and place that word at the end of the comment line above.

0age (changes requested):

reference/lib/ReferenceOrderCombiner.sol
0xBeans: Do you mean this?

```
                    // increment total filtered Executions.
                    // starts at 0, infeasible to increment > 2^256 Executions.
```

reference/lib/ReferenceOrderCombiner.sol
0xBeans: Done.

reference/lib/ReferenceOrderCombiner.sol
0xBeans: Changed comment to `// Executions start at 0, infeasible to increment > 2^256.` and moved above unchecked{} block and capitlized `increments` as that seems to be the standard among the entire code base.

0xBeans:
@0age Capitalized `increment` comment as the entire codebase seems to follow that structure

Screen Shot 2022-06-06 at 6 07 27 PM

hrkrshnn:
I assumed adding `unchecked` into ReferenceOrdercombiner was against the design--keeping Reference without any optimizations. If we should be pedantic about it, this change should be reverted. cc @0age

PR #351, opened 2022-06-06
Fix edge case assertion for FulfillAdvancedOrder fuzz tests

jameswenzel:
## Motivation
When `args.recipient` is `alice`, the `balanceOf` assertion will fail.

## Solution
Conditionally assert either numer or denom balance

### DIFF
--- a/test/foundry/FulfillAdvancedOrder.t.sol
+++ b/test/foundry/FulfillAdvancedOrder.t.sol
@@ -107,24 +107,24 @@ contract FulfillAdvancedOrder is BaseOrderTest {
     }
 
     function testAdvancedPartialAscendingOfferAmount1155(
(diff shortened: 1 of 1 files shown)


test/foundry/FulfillAdvancedOrder.t.sol
```diff
@@ -554,8 +549,11 @@ contract FulfillAdvancedOrder is BaseOrderTest {
             assertFalse(isCancelled);
             assertEq(totalFilled, context.args.numer);
             assertEq(totalSize, context.args.denom);
+            // if recipient is alice (offerer), final balance will be number minted, ie, denom
```
hrkrshnn, line 552: "number minted"?

PR #353, opened 2022-06-06
Added unchecked to FulfillmentApplier

dabushori:
## Motivation

## Solution


0age:
Actually — can't the whole conditional statement be unchecked here?

0age:
Closing as I think this change made it into the codebase via another route. Feel free to open another if there's a way to improve on it from here or if you still need to add your name to `CONTRIBUTORS.md`

PR #354, opened 2022-06-06
Added unchecked to OrderCombiner

dabushori:
## Motivation

## Solution

### DIFF
--- a/contracts/lib/OrderCombiner.sol
+++ b/contracts/lib/OrderCombiner.sol
@@ -232,8 +232,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                 // Otherwise, track the order hash in question.
                 orderHashes[i] = orderHash;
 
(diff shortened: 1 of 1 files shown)


d1ll0n (approved):
Good catch. For reference, it can't be 0 because of https://github.com/ProjectOpenSea/seaport/blob/14a6c0f186522e25eea10d4e4de4ba98db445ea9/contracts/lib/OrderCombiner.sol#L193

dabushori:
> All of this is already in an unchecked block starting at line 186? Aren't you stacking `unchecked` here, so it has no impact? I like the additional comment to `maximumFulfilled--` here though.

Your'e right, my mistake!

0age:
All tests failing due to nested unchecked, can you address?

0age:
Closing as all the tests are red due to nested unchecked, feel free to reopen with the additional comment though!

PR #355, opened 2022-06-06
Added unchecked to OrderFulfiller

dabushori:
## Motivation

## Solution

### DIFF
--- a/contracts/lib/OrderFulfiller.sol
+++ b/contracts/lib/OrderFulfiller.sol
@@ -157,10 +157,13 @@ contract OrderFulfiller is
         bytes32 fulfillerConduitKey,
         address recipient
     ) internal {
-        // Derive order duration, time elapsed, and time remaining.
-        uint256 duration = orderParameters.endTime - orderParameters.startTime;
-        uint256 elapsed = block.timestamp - orderParameters.startTime;
-        uint256 remaining = duration - elapsed;
+        // Skip underflow checks as we know orderParameters.startTime <= block.timestamp < orderParameters.endTime
+        unchecked {
+            // Derive order duration, time elapsed, and time remaining.
+            uint256 duration = orderParameters.endTime - orderParameters.startTime;
+            uint256 elapsed = block.timestamp - orderParameters.startTime;
+            uint256 remaining = duration - elapsed;
+        }
 
         // Put ether value supplied by the caller on the stack.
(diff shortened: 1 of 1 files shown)


contracts/lib/OrderFulfiller.sol
```diff
@@ -157,10 +157,13 @@ contract OrderFulfiller is
         bytes32 fulfillerConduitKey,
         address recipient
     ) internal {
-        // Derive order duration, time elapsed, and time remaining.
-        uint256 duration = orderParameters.endTime - orderParameters.startTime;
-        uint256 elapsed = block.timestamp - orderParameters.startTime;
-        uint256 remaining = duration - elapsed;
+        // Skip underflow checks as we know orderParameters.startTime <= block.timestamp < orderParameters.endTime
```
0age, line 160: Two comments:
- break comment up so that 80 character line limit is not exceeded
- avoid "we" in favor of declarative comments (basically remove "we know")

0age (changes requested):

0age:
looks like you need to declare the variables outside of the unchecked scope and run the linter!

contracts/lib/OrderFulfiller.sol
```diff
@@ -157,10 +157,16 @@ contract OrderFulfiller is
...
         address recipient
     ) internal {
-        // Derive order duration, time elapsed, and time remaining.
-        uint256 duration = orderParameters.endTime - orderParameters.startTime;
-        uint256 elapsed = block.timestamp - orderParameters.startTime;
-        uint256 remaining = duration - elapsed;
+        uint256 duration;
+        uint256 elapsed;
+        uint256 remaining;
+        // Skip underflow checks as startTime <= block.timestamp < endTime
```
MrToph, line 163: to me, it's not immediately obvious that this is true and requires analysing all calling contexts.

dabushori: > to me, it's not immediately obvious that this is true and requires analysing all calling contexts.

Your'e right, but the only call to this function is after verifying these conditions are true

MrToph: It would be nice if the comment was more clear about 1) where this validation happens and 2) mention all the callers and that they do indeed go through this validation step.
Otherwise, when this code is maintained in the future it's easy to do mistakes if another function calls this.

PR #358, opened 2022-06-06
Use calldata ptr constant

laurenceday:
## Motivation

The `transferDataSize` parameter in `__performERC1155BatchTransfers` of `contracts/lib/TokenTransferrer.sol` was reusing a constant `BatchTransfer1155Params_data_length_basePtr` to determine the calldata size for the call to transfer.

## Solution

Simple: introduce a new constant in `TokenTransferrerConstants.sol` called `BatchTransfer1155Params_calldata_baseSize` and use that.


d1ll0n (approved):
Good catch, thanks!

PR #363, opened 2022-06-06
Remove `gt` for `gcd` loop boolean condition

Vectorized:
## Motivation

Ain't much but it's honest work.

We don't need a `gt(_b, 0)` check for `gcd`, since all values are uint256.  

The zeroness of `_b` will suffice as the boolean condition for the loop.

I tried on Remix, the compiler doesn't optimize that one out. 

## Solution

Remove the `gt` check.


axic:
> I tried on Remix, the compiler doesn't optimize that one out.

Are you sure? There is an optimiser rule for this in the compiler. Did you had the optimiser turned on?

Vectorized:
@axic opps. It actually does gets optimized away.

PR #370, opened 2022-06-07
Remove Murky Dependency

dmfxyz:
## Motivation
The [Murky](https://github.com/dmfxyz/murky) dependency looks like it was used in an earlier version of the testing suite, but is no longer used. Unused dependencies should not be in the code base (even if I have a personal attachment to them 😄 ).

## Solution
As much as it pains me to do so as it's my own library (and I'm honored someone thought to use it in the first place), the solution is to remove the library. Maybe it will come back later in the future for additional tests in CriteriaResolution.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,3 @@
 [submodule "lib/solmate"]
 	path = lib/solmate
 	url = https://github.com/rari-capital/solmate
-[submodule "lib/murky"]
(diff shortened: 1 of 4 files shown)


jameswenzel:
Resolved with #371 🤠

dmfxyz:
@jameswenzel wow! nevermind then :)

PR #376, opened 2022-06-07
Reword and fix grammar

DanielGelfand:
## Motivation

To make it clearer when the error is thrown and make it consistent with other comments.

## Solution

Fix grammar `an -> a` and reword to make it clear that it is a fulfillment that does not declare at least one component rather than the orders.


DanielGelfand:
> Can you address the lint warning (line length over 80 characters)?

Fixed

PR #377, opened 2022-06-07
minor gas optimization + loop incremention consistency

wpoulin:
Improved loop incrementation consistency + minor gas optimization


0age:
Can you resolve the merge conflicts?

test/foundry/NonReentrant.t.sol
```diff
@@ -56,8 +56,8 @@ contract NonReentrantTest is BaseOrderTest {
     }
 
     function testNonReentrant() public {
-        for (uint256 i; i < 7; i++) {
-            for (uint256 j; j < 10; j++) {
+        for (uint256 i; i < 7; ++i) {
```
axic, line 59: Which was more prevalent post or pre increment? This PR seems to change it at so many places I wonder if the new style was prevalent 😅

axic: There is also a claim about gas optimization, I only see the `i++` -> `++i` changes, is that really making any changes for gas?

wpoulin: I did update a few `var++ `to `++var` yesterday night, but this morning a bunch of PRs merged main and the gas optimization changes I made were already on master.

PR #380, opened 2022-06-07
Update `CONTRIBUTORS.md` and Assembly cast to uint256

rotcivegaf:
## Motivation

I participated in the code4rena audit contest
And save gas when comparing addresses

## Solution

In an address comparison, cast both addresses to uin256 and compare these
This saves between 13 and 18 gas by comparison


transmissions11:
does this really save gas? oof if so

0age:
Can you run `yarn lint:fix` and commit the changes?

rotcivegaf:
I always forget the lint...
Done

rotcivegaf:
> does this really save gas? oof if so

Learned from solmate :shushing_face:

transmissions11:
> > does this really save gas? oof if so
> 
> Learned from solmate 🤫

When have I done something like this?

0age:
I'm also skeptical of this being value-additive; would need to see some benchmarks of gas usage from before and after given the overhead of the internal function call before making this change, particularly as it damages readability a good deal and also introduces a danger of dirtied bits becoming a factor in the comparisons when they shouldn't be. The change should also be removed from the conduit controller as gas optimization is not the core objective of that contract (it strongly favors clarity of purpose as it is not used frequently and handles highly sensitive operations).

rotcivegaf:
> > > does this really save gas? oof if so
> > 
> > 
> > Learned from solmate shushing_face
> 
> When have I done something like this?

My mistake, I saw some PR in the solmate repository.

0age:
Closing this as the scope is too broad; if there are targeted areas where you think this trick could be applied and well documented, particularly as part of "hot paths" that will spend a lot of gas and so merit optimization, open a new PR with them!

PR #381, opened 2022-06-07
Update Contributors and Optimize a Division

Zer0dot:
## Motivation

I just participated in the C4 contest a little bit, found two microscopic optimizations that reduced gas usage by infinitesimally small amounts. This PR includes the more obvious of the two optimizations, as the other one (checking the success condition and reverting in the `else` block) is a little more opinionated and less necessary. Either way, it's in the C4 report!

It ain't much, but it's honest work

But hey, it's something.


contracts/lib/BasicOrderFulfiller.sol
```diff
@@ -83,7 +83,7 @@ contract BasicOrderFulfiller is OrderValidator {
             orderType := and(calldataload(BasicOrder_basicOrderType_cdPtr), 3)
 
             // Divide basicOrderType by four to derive the route.
-            route := div(calldataload(BasicOrder_basicOrderType_cdPtr), 4)
+            route := shr(2, calldataload(BasicOrder_basicOrderType_cdPtr))
```
transmissions11, line 86: 99% sure the yul optimizer can do this on its own?

0age: may as well be explicit about it 🤷

Zer0dot: > 99% sure the yul optimizer can do this on its own?

Saved gas in my test at least

PR #383, opened 2022-06-07
Small optimizations

azhar0406:
## Motivation

Minor gas optimisations with assembly based odd or even checker and some unchecked for overflow.

### DIFF
--- a/reference/lib/ReferenceOrderValidator.sol
+++ b/reference/lib/ReferenceOrderValidator.sol
@@ -343,6 +343,8 @@ contract ReferenceOrderValidator is
             // Update the order status as not valid and cancelled.
             _orderStatus[orderHash].isValidated = false;
             _orderStatus[orderHash].isCancelled = true;
+            
+            
 
             // Emit an event signifying that the order has been cancelled.
             emit OrderCancelled(orderHash, offerer, zone);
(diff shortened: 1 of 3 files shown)


reference/lib/ReferenceOrderValidator.sol
```diff
@@ -407,6 +409,9 @@ contract ReferenceOrderValidator is
 
                 // Update order status to mark the order as valid.
                 _orderStatus[orderHash].isValidated = true;
+                
+                // check i to prevents overflow.
+                unchecked {++i}
```
MrToph, line 414: There's a bug here where you only increment `i` if the order has not already been validated. So if I pass an `orders` array where `orders[i]` has been validated before, I'll get to an infinite loop and the item does not get skipped.

I also think that they have a rule of no `unchecked` and `assembly` in the reference contracts.

azhar0406: Ohh yes, you are correct, will update it. Moreover, for the reference contracts rule, I didn't know about it. Thanks for letting me know.

0age:
Looks like lint is failing — run `yarn lint:fix` before committing — and there are also some merge conflicts. If the change is small enough you may want to just open a new PR and close this one.

PR #388, opened 2022-06-07
Fix AlmostOneWord from 0x19 to 0x1f = 31

leonardoalt:
Assuming that the comment under https://github.com/ProjectOpenSea/seaport/pull/311/files#diff-39e723c7a1e430233f5f7956e4e0f6d23f77fb7e3f1a11e145f53e215f3610fdR36 is correct, this fixes it.
It should be 31 because the changes in that PR were made to make sure that data rounds up in number of words, which is computed by `(Size + 31) / 32`.


leonardoalt:
It would also be good to write tests that fail if this number is wrong since this was not caught.

0age:
Nice catch! Tests that trip these paths are _very_ tricky to write as they’re so sensitive to gas usage 😅

PR #392, opened 2022-06-07
Update Consideration.sol

AyiSoli:
Updated inline documentation readability.


0age (changes requested):
One section that I think is better as-is; also be sure to run `yarn lint:fix` and `yarn lint:check` before committing!

0age:
Closing this for now, feel free to reopen with one where linters pass and comment's addressed!

PR #393, opened 2022-06-07
Optimization of verify()

gpersoon:
Optimization of verify() by removing branches with multiplications (as is frequently done in other parts of the code).

## Motivation
To save some gas in the Merkleproof functions.

## Solution
As suggested in the Code4rena Seaport competition and further optimized by Rari-Capital
https://github.com/Rari-Capital/solmate/blob/v7/src/utils/MerkleProof.sol


0age:
Mind running `yarn lint:fix` and committing the changes?

PR #395, opened 2022-06-07
Optimize orderStatus as storage pointer

0xPhaze:
Optimize orderStatus to be a storage pointer instead of copying to memory in order to save gas.

## Motivation

orderStatus is mainly being passed around in memory whereas it can directly
be read from storage, reducing gas costs.
Additionally, by caching the storage pointer some gas can be saved.

## Solution

Change
```solidity
OrderStatus memory orderStatus = _orderStatus[orderHash];
```
to
```solidity
OrderStatus storage orderStatus  = _orderStatus[orderHash];
```

Note: also saw that `FulfillmentComponent memory targetComponent` could be changed to `calldata`
And should probably adapt reference implementation as well..


0age:
There's a lot of potentially valuable stuff in this PR, but there are also sections that may actually increase gas (for example, `_verifyOrderStatus` takes `orderHash` as an argument and so this might actually trigger reading it from memory unless the memory declaration on the argument handles the read upfront, need to validate that). Can you break this PR up into a few smaller ones so we can evaluate each change in isolation?

contracts/lib/FulfillmentApplier.sol
```diff
@@ -91,7 +91,7 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
         // If total consideration amount exceeds the offer amount...
         if (considerationItem.amount > execution.item.amount) {
             // Retrieve the first consideration component from the fulfillment.
-            FulfillmentComponent memory targetComponent = (
+            FulfillmentComponent calldata targetComponent = (
```
axic, line 94: These aren't storage changes. Perhaps a separate PR is better.

0xPhaze:
Alright, focused this on only the storage optimization.
`_verifyOrderStatus` only uses `orderHash` for revert messages. Not sure what exactly you mean by "trigger reading it from memory". I also changed the parameter type to `storage` in `_verifyOrderStatus`.

contracts/conduit/ConduitController.sol
```diff
@@ -95,11 +95,13 @@ contract ConduitController is ConduitControllerInterface {
         // Deploy the conduit via CREATE2 using the conduit key as the salt.
         new Conduit{ salt: conduitKey }();
 
+        ConduitProperties storage conduitProperties = _conduits[conduit];
```
0age, line 98: Mind adding a comment above this line? Something like "Initialize storage variable referencing conduit properties."

contracts/lib/Verifiers.sol
```diff
@@ -116,14 +116,16 @@ contract Verifiers is Assertions, SignatureVerification {
             return false;
         }
 
+        uint256 orderStatusNumerator = orderStatus.numerator;
```
0age, line 119: Comment here as well? Something like "Read order status numerator from storage and place on stack."

contracts/lib/Verifiers.sol
```diff
@@ -116,14 +116,16 @@ contract Verifiers is Assertions, SignatureVerification {
...
+
         // If the order is not entirely unused...
-        if (orderStatus.numerator != 0) {
+        if (orderStatusNumerator != 0) {
             // ensure the order has not been partially filled when not allowed.
             if (onlyAllowUnused) {
                 // Always revert on partial fills when onlyAllowUnused is true.
                 revert OrderPartiallyFilled(orderHash);
                 // Otherwise, ensure that order has not been entirely filled.
```
0age, line 127: Unrelated to your PR, but could you also move this comment to below the else statement?

0age (changes requested):
Some final nits, otherwise LGTM!

contracts/lib/Verifiers.sol
0xPhaze: Wasn't sure which comment you were referring to here.

PR #404, opened 2022-06-07
Use correct offset variable to reference objects in assembly

twojoy0:
## Motivation

Use of generic offset variables such as `TwoWords` may be confusing in assembly code(and may also change in the future). To improve readability use specific offset variables to reference objects in assembly code.

## Solution

Replace generic offset variable such as `TwoWords` with specific one eg. `OrderParameters_offer_head_offset` wherever possible


0age:
Mind running `yarn lint:fix` and committing that change?

0age:
looks like #407 beat you to the punch on this one — feel free to just open a PR adding yourself as a contributor though!

PR #406, opened 2022-06-07
Minor optimization for testing `v` is valid for signature verification.

Vectorized:
## Motivation

Sorry, my last contribution was actually a psyops that didn't work. 

So I gotta do something that hopefully won't get optimized away by the compiler.

Copied from https://github.com/Rari-Capital/solmate/pull/247/files

Please feel free to copy more tricks from the solmate PR.

## Solution

Use some assembly to test if `v` is either 27 or 28, and take the negation of the result.

### DIFF
--- a/contracts/lib/SignatureVerification.sol
+++ b/contracts/lib/SignatureVerification.sol
@@ -77,9 +77,9 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
                 v := byte(0, mload(add(signature, ThreeWords)))
 
                 // Whether v is 27 or 28.
-                // The magic number has the 27th and 28th bytes
+                // The magic constant has the 27th and 28th bytes
                 // counting from the most significant byte set to 1.
-                vIsValid := byte(v, 0x0101000000)
+                vIsValid := byte(v, ECDSA_twentySeventhAndTwentyEighthBytesSet)
             }
 
             // Ensure v value is properly formatted.
--- a/contracts/lib/SignatureVerification.sol
+++ b/contracts/lib/SignatureVerification.sol
@@ -63,7 +63,7 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
         } else if (signature.length == 65) {
             // Whether v value is properly formatted.
             bool vIsValid;
-            
+
             // If signature is 65 bytes, parse as a standard signature (r+s+v).
             // Read each parameter directly from the signature's memory region.
             assembly {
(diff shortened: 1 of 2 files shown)


0age:
Can you also run `yarn lint:fix` and commit those changes?

contracts/lib/SignatureVerification.sol
```diff
@@ -72,10 +75,15 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
 
                 // Place final byte on the stack at v.
                 v := byte(0, mload(add(signature, ThreeWords)))
+
+                // Whether v is 27 or 28.
+                // The magic number has the 27th and 28th bytes
+                // counting from the most significant byte set to 1.
+                vIsValid := byte(v, 0x0101000000)
```
axic, line 82: So this supposedly works by seeing if byte 27 or 28 is set in the value, but the confusing part is that `byte` works "left aligned" while the literal is "right aligned".

Should have the full 256 bits as a constant for readability, and I'd argue against moving this into a constants file as that reduces readability.

0age: I do think that it's an improvement to readability to have the named constant, as it answers the question of "what is this value" in a succinct way (yes, if you're already familiar with the EVM at a deep level and "know" what the value is then it's useful to have it inlined but most readers will not have that deep familiarity). It's also consistent with the rest of the codebase.

0age: I mean `TwentySeventhAndTwentyEighthBytesSet` is pretty on the nose and can be evaluated for accuracy in an isolated context...

0age: (or something to that effect)

0age: I hear you that now a reviewer has to go and look up that constant to ensure that it's correct, but the same can be said for masks and offsets

contracts/lib/SignatureVerification.sol
```diff
@@ -72,10 +75,15 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
 
                 // Place final byte on the stack at v.
                 v := byte(0, mload(add(signature, ThreeWords)))
+
+                // Whether v is 27 or 28.
+                // The magic constant has the 27th and 28th bytes
+                // counting from the most significant byte set to 1.
+                vIsValid := byte(v, ECDSA_twentySeventhAndTwentyEighthBytesSet)
```
axic, line 82: You may as well just flip this entire thing into `vIsInvalid`, flipping all the bytes, and getting rid of the `iszero` (aka `!`).

At least then the saving is reasonable (including pushes: 6 gas in total vs. 18 gas for `iszero(or(eq(v, 27), eq(v, 28)))`, otherwise this is borderline making it more complex for not much savings.

axic: Without this currently it is 9 vs 18 gas, which I'd argue is not much of a saving compared to the complexity.

0age (approved):
LGTM! @axic you agree?

axic:
I actually thought one could invert the constant and therefore remove the need for `iszero` altogether, but that has a problem: `byte` returns 0 on out-of-bounds access. So yeah, probably the current version is the optimisations. Someone else should also check it though 😅

d1ll0n (approved):
This is really clever, thank you!

PR #407, opened 2022-06-07
Replace generic constant by OrderParameters_offer_head_offset

hrkrshnn:
TwoWords (0x40) is a generic constant. There is a more appropriate constant
`OrderParameters_offer_head_offset` with the same value here. This constant is already being used in line 181.

For verification:
1. `TwoWords`: https://github.com/ProjectOpenSea/seaport/blob/c084b88f55753cff187efe403c3e2d0117f8286d/contracts/lib/ConsiderationConstants.sol#L100
2. `OrderParameters_offer_head_offset`: https://github.com/ProjectOpenSea/seaport/blob/c084b88f55753cff187efe403c3e2d0117f8286d/contracts/lib/ConsiderationConstants.sol#L89


0age:
mind running `yarn lint:fix` and committing that change?

PR #412, opened 2022-06-07
Use numerators instead of filledNumerator for consistency

axic:


contracts/lib/OrderValidator.sol
```diff
@@ -248,20 +248,19 @@ contract OrderValidator is Executor, ZoneInteraction {
...
-                        gcd(filledNumerator, denominator)
+                        gcd(numerator, denominator)
                     )
 
                     // Note: this may not be necessary — need to validate.
                     let safeScaleDown := add(scaleDown, iszero(scaleDown))
 
                     // Scale all fractional values down by gcd.
                     numerator := div(numerator, safeScaleDown)
-                    filledNumerator := div(filledNumerator, safeScaleDown)
```
axic, line 259: This needs to be verified, but there doesn't seem to be any reason for it.

contracts/lib/OrderValidator.sol
```diff
@@ -280,7 +279,7 @@ contract OrderValidator is Executor, ZoneInteraction {
                 // Update order status and fill amount, packing struct values.
                 _orderStatus[orderHash].isValidated = true;
                 _orderStatus[orderHash].isCancelled = false;
-                _orderStatus[orderHash].numerator = uint120(filledNumerator);
+                _orderStatus[orderHash].numerator = uint120(numerator);
                 _orderStatus[orderHash].denominator = uint120(denominator);
             }
```
axic, line 284: After this PR these two blocks could be moved outside of the if statement and reduced to a single one, as they are identical.

contracts/lib/OrderValidator.sol
```diff
@@ -224,13 +224,13 @@ contract OrderValidator is Executor, ZoneInteraction {
...
 
             // Increment the filled numerator by the new numerator.
-            filledNumerator += numerator;
+            numerator += filledNumerator;
```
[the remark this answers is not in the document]
axic: Note that all variables are `uint256`.

0age:
So in this case the `numerator` refers to the amount that's currently being fulfilled, while `filledNumerator` refers to the _total_ amount that will have been filled. By way of example, if you have an order that's 1/3 filled, then you fulfill another 1/3, then `numerator` will be 1, `filledNumerator` will be 2, and `denominator` will be 3 (and items will all be scaled down to 1/3 of their original amounts during fulfillment, not 2/3rds). Agree that the naming is not great, particularly as the section above uses `filledAmount` as a representation of the amount being filled (and it's not until the numerator is added back that it represents the whole filled amount).

axic:
> So in this case the numerator refers to the amount that's currently being fulfilled, while filledNumerator refers to the total amount that will have been filled.

I'd say this was a test for you or the tests to catch a "convincing" PR. It at least failed on tests.

Truth is I missed one line 😅

0age:
Closing this one, we can reopen a new one if there's a good way to improve on this section

PR #416, opened 2022-06-07
Bump solidity >=0.8.13

Chomtana:
## Motivation

Currently there are many files with solidity >=0.8.7

## Solution

Find all and replace >=0.8.7 with >=0.8.13

I have already submit my name to contributor in https://github.com/ProjectOpenSea/seaport/pull/348

### DIFF
--- a/config/.solhint.json
+++ b/config/.solhint.json
@@ -1,7 +1,7 @@
 {
   "extends": "solhint:all",
   "rules": {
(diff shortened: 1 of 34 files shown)


0age:
Aaah, this is intended; all of the reference contracts are built using 0.8.7 as it was recommended as a good, battle-tested version of the solidity compiler to compare against!

PR #419, opened 2022-06-07
fix error in SignatureVerification natspec

0age:


contracts/lib/SignatureVerification.sol
```diff
@@ -22,7 +22,7 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
      * @dev Internal view function to verify the signature of an order. An
      *      ERC-1271 fallback will be attempted if either the signature length
      *      is not 32 or 33 bytes or if the recovered signer does not match the
```
MrToph, line 24: still wrong here

Thro77le: also missed `ReferenceVerifiers.sol`

Thro77le: also missed in `ReferenceSignatureVerification.sol`

PR #425, opened 2022-06-07
Add contributor and some small stuff

0xsanson:
Some findings for CodeArena contest (for example test/findings/AdditionalRecipientsOffByOne is based on my [gist here](https://gist.github.com/0xsanson/e87e5fe26665c6cecaef2d9c4b0d53f4)).

Also made a couple of fixes :)


0age:
Can you run the linter? `yarn lint:fix`

0xsanson:
sorry, should be fixed

d1ll0n (approved):
Good catch on that being redundant half the time

contracts/lib/FulfillmentApplier.sol
```diff
@@ -122,10 +122,12 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
                     .startAmount = (execution.item.amount -
                     considerationItem.amount);
             }
+
+            // Reuse consideration amount.
```
0age, line 126: One more nit here: since we're now more consistent across treatment of these two conditionals, can we update this comment to be more inline with the other one, i.e. `// Reduce total offer amount to equal the consideration amount.`

0age (changes requested):
One small comment change, otherwise LGTM!

PR #426, opened 2022-06-07
standardize array length caching

zaqk:
## Motivation
I noticed there were a few inconsistencies in the way seaport caches array length.  If this PR is accepted I will update other areas where array length is not being cached before a for loop.

## Solution
Read the length of an array and place the length on the stack before looping the array.

Thanks!

### DIFF
--- a/contracts/lib/OrderCombiner.sol
+++ b/contracts/lib/OrderCombiner.sol
@@ -250,8 +250,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                 // Retrieve array of offer items for the order in question.
                 OfferItem[] memory offer = advancedOrder.parameters.offer;
 
+                // Read length of offer array and place on the stack.
+                uint256 totalOffers = offer.length;
+
                 // Iterate over each offer item on the order.
-                for (uint256 j = 0; j < offer.length; ++j) {
+                for (uint256 j = 0; j < totalOffers; ++j) {
                     // Retrieve the offer item.
                     OfferItem memory offerItem = offer[j];
 
@@ -294,8 +297,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                     advancedOrder.parameters.consideration
                 );
 
+                // Read length of consideration array and place on the stack.
+                uint256 totalConsiderations = consideration.length;
+
                 // Iterate over each consideration item on the order.
-                for (uint256 j = 0; j < consideration.length; ++j) {
+                for (uint256 j = 0; j < totalConsiderations; ++j) {
                     // Retrieve the consideration item.
                     ConsiderationItem memory considerationItem = (
                         consideration[j]
@@ -571,6 +577,9 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
         // Retrieve the length of the advanced orders array and place on stack.
         uint256 totalOrders = advancedOrders.length;
 
+        // Retrieve the length of the executions array and place on stack.
+        uint256 totalExecutions = executions.length;
+
         // Initialize array for tracking available orders.
         availableOrders = new bool[](totalOrders);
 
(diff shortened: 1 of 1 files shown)


contracts/lib/OrderCombiner.sol
```diff
@@ -250,8 +250,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                 // Retrieve array of offer items for the order in question.
                 OfferItem[] memory offer = advancedOrder.parameters.offer;
 
+                // Read length of offer array and place on the stack.
+                uint256 totalOffers = offer.length;
```
0age, line 254: Nit: the offer is an array of offer items, so there's no plural of it — rename to `totalOfferItems`

contracts/lib/OrderCombiner.sol
```diff
@@ -294,8 +297,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                     advancedOrder.parameters.consideration
                 );
 
+                // Read length of consideration array and place on the stack.
+                uint256 totalConsiderations = consideration.length;
+
                 // Iterate over each consideration item on the order.
-                for (uint256 j = 0; j < consideration.length; ++j) {
+                for (uint256 j = 0; j < totalConsiderations; ++j) {
```
0age, line 304: Same idea here: `totalConsiderationItems`

contracts/lib/OrderCombiner.sol
```diff
@@ -597,8 +606,11 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
                     advancedOrder.parameters.consideration
                 );
 
+                // Read length of consideration array and place on the stack.
+                uint256 totalConsiderations = consideration.length;
```
0age, line 610: Same nit as above here

0age (changes requested):
Some naming nits, but otherwise LGTM! Heads-up that a number of these inconsistencies originate from stack-too-deep errors; and that's not just on the main runs but also on code coverage which does not use viaIR.

contracts/lib/OrderCombiner.sol
zaqk: ah got it that makes sense. let me update. thanks.

btw the stack-too-deep errors should be better handled in 0.8.14 right? so is it no longer an issue?

zaqk: can't find a source at the moment so disregard. will tag you on this PR with a source if/when i find it.

0age:
Let's merge this one in assuming tests come back green and then tackle other instances in separate PRs? Also, want to add your name to `CONTRIBUTORS.md`?

zaqk:
> Let's merge this one in assuming tests come back green and then tackle other instances in separate PRs? Also, want to add your name to `CONTRIBUTORS.md`?

Yes would love to. thanks.

0age:
One _tiny_ additional nit — can you move `totalExecutions` so that it's declared just before the array where it's used?

zaqk:
> One _tiny_ additional nit — can you move `totalExecutions` so that it's declared just before the array where it's used?

yes i considered this too. wasn't sure which would be better.

contracts/lib/OrderCombiner.sol
```diff
@@ -625,6 +622,9 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
         // Put ether value supplied by the caller on the stack.
         uint256 etherRemaining = msg.value;
 
+        // Retrieve the length of the executions array and place on stack.
+        uint256 totalExecutions = executions.length;
+
```
zaqk, line 627: I put this above the accumulator variable based off the accumulator variable comment. Let me know if im misreading this. currently running tests locally to see if it will pass while the variable is allocated after the accumulator variable.

0age: oooh that only corrupts new memory declared after the accumulator is instantiated; you can / should just use it right before the iterator!

0age: i.e. executions has already been allocated in memory

zaqk: Ooooooh got it. that makes sense thanks for explaining.

zaqk:
@0age just following up on this. There is one other function that does not cache the array length (https://github.com/ProjectOpenSea/seaport/blob/main/contracts/lib/OrderFulfiller.sol#L153). But adding any other parameters here causes a stack-too-deep error unfortunately. It's ~10 gas every loop regardless of array length so not a huge deal. In total it's about 20 extra gas for the entire function (the function has two uncached loops).

PR #430, opened 2022-06-07
index `ChannelUpdated` event parameter `channel`

berndartmueller:
## Motivation

Indexed event parameters are stored in the topics part of the log instead of the data part, which allows for faster indexing/querying because of the use of bloom filters for topics. Up to three parameters in every event can be indexed. Indexed event parameters allow for faster and more efficient/effective event lookups.

Most events already have indexed parameters, therefore it's a good idea to also index the remaining events.

## Solution

Add `indexed` keyword to index event `address` parameter

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -47,3 +47,4 @@ okkothejawa                    |
 FlameHorizon                   |
 vdrg                           |
 Ellahi                         | `ellahi.eth`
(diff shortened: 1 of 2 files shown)


0age:
LGTM! Want to also add yourself to `CONTRIBUTORS.md`?

berndartmueller:
@0age Yes, I would love to. Will add myself.

Thanks!

PR #435, opened 2022-06-07
Add tests for fulfill basic order with invalid length for additional recipients

stuckinaboot:
## Motivation

We have no tests for invalid length of additional recipients in the calldata for `FulfillBasicOrder`.

## Solution

Add fuzz tests for invalid length of additional recipients in the calldata for `FulfillBasicOrder`, where the tests check both
1. `fulfillBasicOrder` does not revert when additional recipients length is not modified
2. `fulfillBasicOrder` does revert when additional recipients length is modified (made smaller than the actual additional recipients length)

## Testing
`forge test --match-contract FulfillBasicOrderTest` all tests pass
`forge test --match-contract NonReentrantTest` all tests pass


0age (approved):
LGTM!

PR #440, opened 2022-06-08
gas: add unchecked for ConduitController#updateChannel

mingwatch:
## Motivation

For the arithmetic operations that will never over/underflow, using the unchecked directive (Solidity v0.8 has default overflow/underflow checks) can save some gas from the unnecessary internal over/underflow checks.

In `ConduitController.sol#updateChannel()`, when `channelPreviouslyOpen` is `true` means that `channelIndexPlusOne != 0`, and `channelIndexPlusOne` is `uint256`, so `channelIndexPlusOne` must >= 1, then `channelIndexPlusOne - 1` will never underflow.

## Solution

add `unchecked` for that case

### DIFF
--- a/contracts/conduit/ConduitController.sol
+++ b/contracts/conduit/ConduitController.sol
@@ -154,8 +154,11 @@ contract ConduitController is ConduitControllerInterface {
         } else if (!isOpen && channelPreviouslyOpen) {
             // Set a previously open channel as closed via "swap & pop" method.
             // Decrement located index to get the index of the closed channel.
-            uint256 removedChannelIndex = channelIndexPlusOne - 1;
-
+            uint256 removedChannelIndex;
+            // channelPreviouslyOpen already make sure channelIndexPlusOne > 0
+            unchecked {
+                removedChannelIndex = channelIndexPlusOne - 1;
+            }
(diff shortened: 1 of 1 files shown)


contracts/conduit/ConduitController.sol
```diff
@@ -154,8 +154,11 @@ contract ConduitController is ConduitControllerInterface {
         } else if (!isOpen && channelPreviouslyOpen) {
             // Set a previously open channel as closed via "swap & pop" method.
             // Decrement located index to get the index of the closed channel.
-            uint256 removedChannelIndex = channelIndexPlusOne - 1;
-
+            uint256 removedChannelIndex;
+            // channelPreviouslyOpen already ensures channelIndexPlusOne > 0
+            unchecked {
+                removedChannelIndex = channelIndexPlusOne - 1;
```
leonardoalt, line 160: This looks safe

leonardoalt: To clarify: this is safe.
In this branch, `channelPreviouslyOpen` must be true, which means that `channelIndexPlusOne != 0`. Since it has type `uint`, it must be `> 0` therefore safe.

PR #441, opened 2022-06-08
EIP1271 verification for non standard signature with 64 or 65 length

Chomtana:
## Motivation

EIP-1271 Signature may contains 64 or 65 bytes but isn't EIP-2098 signature. But it may be reverted if it isn't EIP-2098 signature.

EIP-1271 Signature that contains 64 or 65 bytes may be custom kind of signature for example it may be keccak256 of some value or merkle tree root.

In the case that 64 or 65 bytes EIP-1271 Signature is not EIP-2098 signature. It will be reverted with `InvalidSignature()` or `BadSignatureV(v)` because it is checked with EIP-2098 verification algorithm first and recoveredSigner may be address(0)

Assume current signature is a valid EIP-1271 signature (calling `isValidSignature` on signer return correct magicValue).
(Using EIP1271WalletSpecial that I have hardcoded an invalid EIP-2098 signature as a valid one)

```
      if (signature.length == 64) {
          // Declare temporary vs that will be decomposed into s and v.
          bytes32 vs;

          (r, vs) = abi.decode(signature, (bytes32, bytes32));

          s = vs & EIP2098_allButHighestBitMask;

          v = uint8(uint256(vs >> 255)) + 27;
      }
```

This code block will be passed. But ecrecover may return 0

If ecrecover return 0

```
      if (recoveredSigner == address(0)) {
          revert InvalidSignature();
          // Should a signer be recovered, but it doesn't match the signer...
      }
```

It will be reverted with `InvalidSignature()`.

But in fact it should be valid due to assumption above. ( must perform `_assertValidEIP1271Signature(signer, digest, signature);` )

If ecrecover return non-zero address

```
      } else if (recoveredSigner != signer) {
          // Attempt EIP-1271 static call to signer in case it's a contract.
          _assertValidEIP1271Signature(signer, digest, signature);
      }
```

This case is valid as it is calling `_assertValidEIP1271Signature(signer, digest, signature);` as expected

### Case signature length 65

```
      } else if (signature.length == 65) {
          (r, s) = abi.decode(signature, (bytes32, bytes32));
          v = uint8(signature[64]);

          // Ensure v value is properly formatted.
          if (v != 27 && v != 28) {
              revert BadSignatureV(v);
          }
      }
```

In this case, It may be reverted with revert `BadSignatureV(v);` if length is 65 but uint8(signature[64]) is not 27 or 28.

But in fact it should be valid due to assumption above. ( must perform `_assertValidEIP1271Signature(signer, digest, signature);` )

### Other length

Signature of other length is verified directly into signer contract by calling `_assertValidEIP1271Signature(signer, digest, signature);`, so it will be passed.

```
      } else {
          // For all other signature lengths, try verification via EIP-1271.
          // Attempt EIP-1271 static call to signer in case it's a contract.
          _assertValidEIP1271Signature(signer, digest, signature);

          // Return early if the ERC-1271 signature check succeeded.
          return;
      }
```

## Solution

Check isContract on signer first, if it is contract only verify EIP-1271 as contract never sign EIP-2098 under his address.

If signer is an EOA, only check for EIP-2098 as EOA never use EIP-1271.

If signer is an EOA but signature is malformed or recovered to different signer, revert.

### DIFF
--- a/contracts/lib/SignatureVerification.sol
+++ b/contracts/lib/SignatureVerification.sol
@@ -36,6 +36,20 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
         bytes32 digest,
         bytes memory signature
     ) internal view {
+        bool isContract;
+        assembly {
+            isContract := iszero(iszero(extcodesize(signer)))
+        }
+
+        if (isContract) {
+            // For all contracts, try verification via EIP-1271 as EIP-2098 signer never resolved to a contract.
+            // Attempt EIP-1271 static call to signer in case it's a contract.
+            _assertValidEIP1271Signature(signer, digest, signature);
+
+            // Return early if the ERC-1271 signature check succeeded.
+            return;
+        }
+
         // Declare r, s, and v signature parameters.
         bytes32 r;
         bytes32 s;
@@ -79,12 +93,8 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
                 revert BadSignatureV(v);
             }
         } else {
-            // For all other signature lengths, try verification via EIP-1271.
-            // Attempt EIP-1271 static call to signer in case it's a contract.
-            _assertValidEIP1271Signature(signer, digest, signature);
-
-            // Return early if the ERC-1271 signature check succeeded.
-            return;
+            // Neither EIP-1271 nor EIP-2098
+            revert InvalidSigner();
         }
 
         // Attempt to recover signer using the digest and signature parameters.
@@ -95,8 +105,8 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
             revert InvalidSignature();
             // Should a signer be recovered, but it doesn't match the signer...
         } else if (recoveredSigner != signer) {
-            // Attempt EIP-1271 static call to signer in case it's a contract.
-            _assertValidEIP1271Signature(signer, digest, signature);
+            // EIP-1271 is checked at the first place
+            revert InvalidSigner();
         }
     }
 
--- a/contracts/test/EIP1271WalletSpecial.sol
+++ b/contracts/test/EIP1271WalletSpecial.sol
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.7;
+
+interface ERC20ApprovalInterface {
+    function approve(address, uint256) external returns (bool);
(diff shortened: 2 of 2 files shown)


Vectorized:
This will lead to increased gas usage. 

The `extcodesize` opcode costs 2600 gas if the address is cold. In comparison, `ecrecover` costs around 3000 gas. 

The current code is using the cheap length comparisons to choose whether to use 1271 or regular ecrecover. 

Also, the majority of cases will likely be non-1271. Doing the length checks and regular ecrecover first will save more gas on average.

Chomtana:
Yes, it cause huge gas usage increase. Developing a try catch version which should not use high gas on majority of cases.

Chomtana:
I have pushed an implementation using assembly technique to store revert error halfway, let EIP-1271 verification run on all cases and merged the error with the EIP-2098 one.

This shouldn't cost any extra gas on EOA EIP-2098 theoretically.

And I have create a use case where EIP-1271 compatible wallet is used as an intermediate wallet for crossing chain with message passing.

That kind of wallet require 65 bytes signature with (commandId, approvalHash, version). Which fail in the old codebase, but now worked.

gpersoon:
@Chomtana your approach is very similar to mine now.
Note: This file also has to be updated `ReferenceSignatureVerification.sol`
I'm closing https://github.com/ProjectOpenSea/seaport/pull/447

0age:
@Chomtana do you intend to work on the reference implementation and failing coverage tests? Let us know!

Chomtana:
> @Chomtana do you intend to work on the reference implementation and failing coverage tests? Let us know!

I will work on it today :)

0age:
Closing as this has been folded up into #475 — be sure to add yourself to `CONTRIBUTORS.md` if you haven't already!

gpersoon:
If `signer == 0`  and  `signature.length` != 27 or 28 then this function could allow the signature.

As a safety precaution: check signer !=0
```solidity 
function _assertValidSignature(address signer, ..., bytes memory signature ) ... {
    address recoveredSigner;
    assembly {
        let signatureLength := mload(signature)
        let lenDiff := sub(ECDSA_MaxLength, signatureLength)
        ...           
        if lt(lenDiff, 2) {    // assume this false
            ...
        }
        ...
        success := eq(signer, recoveredSigner)   // if signer happens to be 0 then success = true
        ...
    }
}
```

PR #447, opened 2022-06-08
First ecrecover is tried and if that fails ERC-1271 is tried

gpersoon:
This is an alternative solution to https://github.com/ProjectOpenSea/seaport/pull/441

## Motivation
The current solution fails with Gnosis safe.
Verifying a the signer is a contract wallet is relative expensive, so this solution tries ecrecover first. 

## Solution
First ecrecover is tried and if that fails ERC-1271 is tried

### DIFF
--- a/contracts/lib/SignatureVerification.sol
+++ b/contracts/lib/SignatureVerification.sol
@@ -33,6 +33,10 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
         bytes32 digest,
         bytes memory signature
     ) internal view {
+        if (signer == 0) {
+           // Make sure to never allow the ecrecover "error" case
+           revert InvalidSigner();
+        }
         address recoveredSigner = _tryEcRecoverSignature(digest, signature);
         // Found a match: recovered address matches the signer
         if (recoveredSigner == signer) return;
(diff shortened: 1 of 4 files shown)


contracts/lib/SignatureVerification.sol
```diff
@@ -36,6 +33,31 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
         bytes32 digest,
         bytes memory signature
     ) internal view {
+        address recoveredSigner = _tryEcRecoverSignature(digest, signature);
```
hrkrshnn, line 36: I think it's better to add a check for `0` here. A bit worried if `signer` can be `0` for the call.

contracts/lib/SignatureVerification.sol
```diff
@@ -83,31 +104,16 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
                     byte(v, ECDSA_twentySeventhAndTwentyEighthBytesSet)
                 )
             }
-
-            // Ensure v value is properly formatted.
+            // Ensure v value is properly formatted. Return invalid value if not
```
axic, line 107:
```suggestion
            // Early abort if v value is out of range to save the cost of ecrecover.
```

axic:
> The current solution fails with Gnosis safe.

See our submission to C4 once released, but basically some signature modes supported by Safe do not work with the current code.

gpersoon:
Closed as https://github.com/ProjectOpenSea/seaport/pull/441 is very similar now and has better error handling.

PR #448, opened 2022-06-08
Gas opt: skip `remaining` calculation when not necessary

naps62:
## Motivation

Critical code paths for order fulfillment are preemptively computing `duration`, `elapsed` and `remaining` timestamps for all orders.
In particular, `remaining` can always be derived from the other two, and it's only actually necessary when `startAmount != endAmount`. For static-price orders, it ends up never actually being used (both `_locateCurrentAmount` and `_applyFraction` branch out and never use it).

This opens the door to a couple of optimizations:
* we can remove this argument from internal calls, and derive it from `duration` and `elapsed` when necessary
* we end up not actually computing it at all when `startAmount == endAmount`, which should be a significant number of orders

`yarn coverage` of the relevant section (before and after shown below) shows gains of 130 to 160 gas for all `fulfill*` and `match*` calls.

**TODO**:
- [x] Update reference implementation as well

main:

```
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAdvancedOrder            ·     100495  ·     209321  ·      160454  ·          126  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAvailableAdvancedOrders  ·     137853  ·     227269  ·      193361  ·           19  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAvailableOrders          ·     171746  ·     226914  ·      205200  ·           13  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillBasicOrder               ·      92295  ·    1628323  ·      665700  ·          160  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillOrder                    ·     101861  ·     212811  ·      173831  ·          105  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  incrementCounter                ·          -  ·          -  ·       47029  ·            6  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  matchAdvancedOrders             ·     204305  ·     271088  ·      252862  ·           67  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  matchOrders                     ·     164525  ·     363989  ·      265784  ·          105  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
```

patched:

```
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAdvancedOrder            ·     100247  ·     209063  ·      160207  ·          126  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAvailableAdvancedOrders  ·     137767  ·     227033  ·      193234  ·           19  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillAvailableOrders          ·     171646  ·     226690  ·      205057  ·           13  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillBasicOrder               ·      92295  ·    1628323  ·      665700  ·          160  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  fulfillOrder                    ·     101613  ·     212553  ·      173575  ·          105  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  incrementCounter                ·          -  ·          -  ·       47029  ·            6  ·          -  │                
······························|··································|·············|·············|··············|···············|··············                
|  Seaport                    ·  matchAdvancedOrders             ·     204109  ·     270892  ·      252636  ·           67  ·          -  │                
······························|··································|·············|·············|··············|···············|··············
```

### DIFF
--- a/contracts/lib/AmountDeriver.sol
+++ b/contracts/lib/AmountDeriver.sol
@@ -24,8 +24,6 @@ contract AmountDeriver is AmountDerivationErrors {
      *
      * @param startAmount The starting amount of the item.
      * @param endAmount   The ending amount of the item.
-     * @param elapsed     The time elapsed since the order's start time.
-     * @param duration    The total duration of the order.
      * @param roundUp     A boolean indicating whether the resultant amount
      *                    should be rounded up or down.
      *
@@ -34,14 +32,19 @@ contract AmountDeriver is AmountDerivationErrors {
     function _locateCurrentAmount(
         uint256 startAmount,
         uint256 endAmount,
-        uint256 elapsed,
-        uint256 duration,
+        uint256 startTime,
+        uint256 endTime,
         bool roundUp
-    ) internal pure returns (uint256) {
+    ) internal view returns (uint256) {
         // Only modify end amount if it doesn't already equal start amount.
         if (startAmount != endAmount) {
             // Leave extra amount to add for rounding at zero (i.e. round down).
             uint256 extraCeiling = 0;
+
+            uint256 duration = endTime - startTime;
+            uint256 elapsed = block.timestamp - startTime;
+            // Derive time remaining until order expires
+            uint256 remaining = duration - elapsed;
 
             // If rounding up, set rounding factor to one less than denominator.
             if (roundUp) {
@@ -50,9 +53,6 @@ contract AmountDeriver is AmountDerivationErrors {
                     extraCeiling = duration - 1;
                 }
             }
-
-            // Derive time remaining until order expires
-            uint256 remaining = duration - elapsed;
 
             // Aggregate new amounts weighted by time with rounding factor.
             // prettier-ignore
@@ -129,8 +129,6 @@ contract AmountDeriver is AmountDerivationErrors {
      * @param numerator       A value indicating the portion of the order that
      *                        should be filled.
      * @param denominator     A value indicating the total size of the order.
-     * @param elapsed         The time elapsed since the order's start time.
-     * @param duration        The total duration of the order.
      * @param roundUp         A boolean indicating whether the resultant
      *                        amount should be rounded up or down.
      *
@@ -141,10 +139,10 @@ contract AmountDeriver is AmountDerivationErrors {
         uint256 endAmount,
         uint256 numerator,
         uint256 denominator,
-        uint256 elapsed,
-        uint256 duration,
+        uint256 startTime,
+        uint256 endTime,
         bool roundUp
-    ) internal pure returns (uint256 amount) {
+    ) internal view returns (uint256 amount) {
         // If start amount equals end amount, apply fraction to end amount.
         if (startAmount == endAmount) {
             // Apply fraction to end amount.
@@ -154,8 +152,8 @@ contract AmountDeriver is AmountDerivationErrors {
             amount = _locateCurrentAmount(
                 _getFraction(numerator, denominator, startAmount),
                 _getFraction(numerator, denominator, endAmount),
-                elapsed,
-                duration,
+                startTime,
+                endTime,
                 roundUp
             );
         }
--- a/contracts/lib/AmountDeriver.sol
+++ b/contracts/lib/AmountDeriver.sol
@@ -142,7 +142,7 @@ contract AmountDeriver is AmountDerivationErrors {
         uint256 numerator,
         uint256 denominator,
         uint256 elapsed,
-        uint256 remaining,
+        // uint256 remaining,
         uint256 duration,
         bool roundUp
     ) internal pure returns (uint256 amount) {
@@ -156,7 +156,8 @@ contract AmountDeriver is AmountDerivationErrors {
                 _getFraction(numerator, denominator, startAmount),
                 _getFraction(numerator, denominator, endAmount),
                 elapsed,
-                remaining,
+                // remaining,
+                duration - elapsed,
                 duration,
                 roundUp
             );
--- a/contracts/lib/AmountDeriver.sol
+++ b/contracts/lib/AmountDeriver.sol
@@ -17,15 +17,15 @@ import "./ConsiderationConstants.sol";
  */
 contract AmountDeriver is AmountDerivationErrors {
     /**
-     * @dev Internal pure function to derive the current amount of a given item
+     * @dev Internal view function to derive the current amount of a given item
      *      based on the current price, the starting price, and the ending
      *      price. If the start and end prices differ, the current price will be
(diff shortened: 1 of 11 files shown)


contracts/lib/AmountDeriver.sol
```diff
@@ -53,6 +51,9 @@ contract AmountDeriver is AmountDerivationErrors {
                 }
             }
 
+            // Derive time remaining until order expires
+            uint256 remaining = duration - elapsed;
```
naps62, line 55: TODO: could this potentially be `unchecked`? Still need to investigate

naps62:
After a better look, this could even be further optimized, by sending directly `startTime` and `endTime` down, and let `_locateCurrentAmount` compute `duration` and `elapsed` as well

contracts/lib/AmountDeriver.sol
leonardoalt: Constraints that are true from c4 review:

1. $startTime \ge block.timestamp_t$
2. $endTime > block.timestamp_t$
3. $endTime > startTime$ (1, 2)
4. $duration = endTime - startTime$
5. $duration > 0$ (3, 4)
6. $elapsed = block.timestamp_w - startTime$
7. $elapsed \ge 0$
8. $block.timestamp = endTime \implies invalid(order)$
9. $elapsed < duration$ (6, 8)

From this I think `unchecked` is fine.

leonardoalt:
For ref, see https://github.com/ProjectOpenSea/seaport/pull/355/files which is doing similar things.

naps62:
Now defering computation of `duration` and `elapsed` as well, as per my previous comment

For now deferred the decision of adding `unchecked`, since it warrants more review, and already generated some comments on #355

contracts/lib/AmountDeriver.sol
```diff
@@ -35,16 +34,24 @@ contract AmountDeriver is AmountDerivationErrors {
...
         // Only modify end amount if it doesn't already equal start amount.
         if (startAmount != endAmount) {
             // Leave extra amount to add for rounding at zero (i.e. round down).
             uint256 extraCeiling = 0;
 
+            // Derive the duration for the order and place it on the stack.
+            uint256 duration = endTime - startTime;
+
+            // Derive time elapsed since the order started & place on stack.
+            uint256 elapsed = block.timestamp - startTime;
```
naps62, line 50: this ended up making the change much bigger than expected, since changing the function from `pure` to `view` and relying on current timestamp force me to switch around a lot of code in the test suite

a better compromise could have been to instead keep the function `pure` and send in an extra `currentTime`. Someone please let me know if that's preferable, and I'll make the change

0age: I think deferring these calculations until necessary makes a lot of sense. If you add a section to the natspec here documenting the conditions asserted in `_verifyTime` you could also make this computation unchecked

contracts/lib/AmountDeriver.sol
```diff
@@ -24,9 +24,8 @@ contract AmountDeriver is AmountDerivationErrors {
      *
      * @param startAmount The starting amount of the item.
      * @param endAmount   The ending amount of the item.
-     * @param elapsed     The time elapsed since the order's start time.
-     * @param remaining   The time left until the order's end time.
-     * @param duration    The total duration of the order.
+     * @param startTime   The starting time of the order
```
0age, line 27: Periods at the end of these comments please

contracts/lib/AmountDeriver.sol
```diff
@@ -128,9 +135,8 @@ contract AmountDeriver is AmountDerivationErrors {
      * @param numerator       A value indicating the portion of the order that
      *                        should be filled.
      * @param denominator     A value indicating the total size of the order.
-     * @param elapsed         The time elapsed since the order's start time.
-     * @param remaining       The time left until the order's end time.
-     * @param duration        The total duration of the order.
+     * @param startTime       The starting time of the order
```
0age, line 138: Same goes for the these and others (use periods at the end)

contracts/lib/OrderCombiner.sol
```diff
@@ -236,14 +236,8 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
...
 
-                // Derive the duration for the order and place it on the stack.
-                uint256 duration = advancedOrder.parameters.endTime - startTime;
-
-                // Derive time elapsed since the order started & place on stack.
-                uint256 elapsed = block.timestamp - startTime;
-
-                // Derive time remaining until order expires and place on stack.
-                uint256 remaining = duration - elapsed;
+                // Place the end for the order on the stack.
```
0age, line 239: End time

reference/lib/ReferenceAmountDeriver.sol
```diff
@@ -35,16 +34,24 @@ contract ReferenceAmountDeriver is AmountDerivationErrors {
...
         uint256 startAmount,
         uint256 endAmount,
-        uint256 elapsed,
-        uint256 remaining,
-        uint256 duration,
+        uint256 startTime,
+        uint256 endTime,
         bool roundUp
-    ) internal pure returns (uint256) {
+    ) internal view returns (uint256) {
```
0age, line 40: Be sure to update the natspec from “Internal pure function…” to “Internal view function…” for the functions that have the new visibility

0age (changes requested):
This is nice! Left some general cleanup notes.

naps62:
@0age fixed all pending comments, and included underflow noticed as requested (actually re-used the notice from #355)

as for the test suite, it was failing to compile due to a "stack too deep error", which I fixed. However, the forge suite is failing for me locally, for reasons I haven't yet understood. It fails on `main` as well though, so I don't think it's a problem here. can we try re-runing the action?

0age:
Rerunning forge tests now!

0age:
While we're working out what's going on with the forge tests, just wanted to also flag an implication of this change: for orders with multiple ascending / descending amount items, these parameters will now be calculated multiple times. It does seem like most items won't have ascending or descending amounts, and the orders that do have an item with ascending or descending amounts usually won't have _more_ than one, but just wanted to flag it and hear what you think.

PR #449, opened 2022-06-08
Remove unnecessary address(0) reverts.

Jon-Becker:
## Motivation

Sometimes, reverting for the 0 address is a waste of gas. It's a check from when the solidity compiler was less strict with calldata lengths.

## Solution

I've removed 2 instances of this check where it might be unnecessary.  

- Lines 62-65 were removed because the initial owner shouldn't ever be the 0 address, unless the caller wants it to be the 0 address, in which case it seems pretty stupid on their part and a waste of gas regardless.

- Lines 205-209 can actually be removed because the check is redundant. If the conduitOwner wants the 0 address to be the newPotentialOwner, there's no real reason to not let that happen. The 0 address can't accept this offer, so ownership won't be transferred to them ever.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -56,3 +56,4 @@ dmfxyz                         | `dmfxyz.eth`
 daltoncoder                    | `dontkillrobots.eth`
 0xf4ce                         | `0xf4ce.eth`
 phaze                          | `phaze.eth`
+Jonathan Becker                | `jbecker.eth`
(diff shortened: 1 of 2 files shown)


Jon-Becker:
This is controversial and can be ignored if you want. I'll look into more substantial contributions soon.

0age:
Closing as we're not gas-constrained on the conduit controller; would love to see a followup PR with another contribution though!

PR #454, opened 2022-06-08
ReferenceSeaport renaming

cupOJoseph:
renames things to seaport in reference folder
continuation of #286 

## Motivation

Reference should be a closer mirror of the main contracts, so it should be named seaport too.

## Solution

renames ReferenceConsideration to ReferenceSeaport and updates all the ref tests.
`yarn test:ref` verified


reference/ReferenceSeaport.sol
```diff
@@ -25,7 +25,7 @@ import { ReferenceOrderCombiner } from "./lib/ReferenceOrderCombiner.sol";
 import { OrderToExecute, AccumulatorStruct } from "./lib/ReferenceConsiderationStructs.sol";
 
 /**
- * @title ReferenceConsideration
+ * @title ReferenceSeaport
```
0age, line 28: This should be a new contract that imports ReferenceConsideration like on the optimized contracts so that we can override the name / name string

0age (changes requested):
One note (should still import ReferenceConsideration and then implement the name override), otherwise LGTM!

reference/ReferenceSeaport.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @dev Internal pure function to retrieve and return the name of this
+     *      contract.
+     *
+     * @return The name of this contract.
+     */
+    function _name() internal pure override returns (string memory) {
+        // Return the name of the contract.
+        assembly {
```
0age, line 25: Since this is the reference implementation we can just return `_nameString()` here and not use assembly!

reference/lib/ReferenceConsideration.sol
```diff
@@ -20,12 +20,12 @@ import {
...
 
-import { ReferenceOrderCombiner } from "./lib/ReferenceOrderCombiner.sol";
+import { ReferenceOrderCombiner } from "./ReferenceOrderCombiner.sol";
 
-import { OrderToExecute, AccumulatorStruct } from "./lib/ReferenceConsiderationStructs.sol";
+import { OrderToExecute, AccumulatorStruct } from "./ReferenceConsiderationStructs.sol";
 
 /**
- * @title ReferenceConsideration
+ * @title ReferenceSeaport
```
0age, line 28: This one should be changed back to ReferenceConsideration

PR #457, opened 2022-06-08
Revert on dirty bits (unused item parameters)

axic:
The goal here is to avoid having orders which look like ERC-721 items with a single byte difference (the `ItemType`), i.e. a `ItemType.NATIVE` order ignores the `token` and `identifier` fields and so it can be used to look like a ERC-721 token (such as BAYC).


0age:
This still needs a lint run, comments, and (importantly) tests to ensure full code coverage; going to merge this in to a new branch so we can finish prep before deciding whether to merge into main!

axic:
> Can you run yarn lint:fix and recommit?

I did but lint was fine on my side for some reason.

PR #458, opened 2022-06-08
perform overflow check on first fill

0age:


contracts/lib/OrderValidator.sol
```diff
@@ -282,6 +283,21 @@ contract OrderValidator is Executor, ZoneInteraction {
                 orderStatus.denominator = uint120(denominator);
             }
         } else {
+            // Use assembly to ensure fractional amounts are below max uint120.
+            assembly {
+                // Check numerator and denominator for uint120 overflow.
+                if or(gt(numerator, MaxUint120), gt(denominator, MaxUint120)) {
```
hrkrshnn, line 289: Is this possible?

leonardoalt: why not?

0age: Yeah on closer inspection, numerator and denominator on the AdvancedOrder struct are already uint120, so casting to uint256 should not change the fact that they cannot exceed uint120:
```solidity
struct AdvancedOrder {
    OrderParameters parameters;
    uint120 numerator;
    uint120 denominator;
    bytes signature;
    bytes extraData;
}
```

PR #459, opened 2022-06-08
Fix to reference Executor itemType

dmitriia:
Seaport C4 Contest, a fix to reference Executor itemType


Thro77le:
Also missed here:
https://github.com/ProjectOpenSea/seaport/blob/main/contracts/lib/Executor.sol#L276
https://github.com/ProjectOpenSea/seaport/blob/main/contracts/lib/Executor.sol#L329
https://github.com/ProjectOpenSea/seaport/blob/main/contracts/lib/Executor.sol#L380

dmitriia:
@0age @Thro77le Added the same for Executor

PR #461, opened 2022-06-08
Add invalid length fulfill order tests

stuckinaboot:
## Motivation

We have no tests for invalid length of consideration items in the calldata for `FulfillOrder`.

## Solution

Add fuzz tests for invalid length of consideration items in the calldata for `FulfillOrder`, where the tests check both
1. `fulfillOrder` does not revert when consideration items length is not modified
2. `fulfillOrder` does revert when consideration items length is modified (made smaller than the actual consideration items length)

Additionally, refactor the similar tests for fulfill basic order and abstract out the common logic between these new tests and those existing tests to live in `BaseOrderTest`, reducing redundant code

## Testing

`forge test --match-test testFulfillOrderRevertInvalidAdditionalRecipientsLength` all tests pass

`forge test --match-contract FulfillOrderTest` all tests pass
`forge test --match-contract FulfillBasicOrderTest` all tests pass


test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +512,96 @@ contract FulfillOrderTest is BaseOrderTest {
         );
     }
 
+    function _performTestFulfillOrderRevertInvalidArrayLength(
```
stuckinaboot, line 515: This will be moved to `BaseOrderTest`

test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +512,96 @@ contract FulfillOrderTest is BaseOrderTest {
...
+            consideration,
+            _order,
+            fulfillOrderCalldata,
+            0x60,
+            amountToSubtractFromConsiderationItemsLength
+        );
+
+        // Ensure consideration items length was properly changed.
+        assertEq(
+            _orderParameters.consideration.length,
```
stuckinaboot, line 599: This check relying on the current _orderParameters length works

test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +512,100 @@ contract FulfillOrderTest is BaseOrderTest {
...
+                itemsLengthFieldOffset,
+                amtToSubtractFromItemsLength
+            );
+        }
+
+        uint256 finalItemsLength = _getLengthAtOffsetInOrderCalldata(
+            fulfillOrderCalldata,
+            itemsLengthFieldOffset
+        );
+        emit log_named_uint("lengt fooh", finalItemsLength);
```
stuckinaboot, line 552: @jameswenzel does this `finalItemsLength` read look reasonable? I'm getting the following emitted: `115792089237316195423570985008687907853269984665640564039457584007913129639935`

test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +512,204 @@ contract FulfillOrderTest is BaseOrderTest {
...
+        (
+            Order memory _order,
+            OrderParameters memory _orderParameters,
+
+        ) = _prepareOrder(1, totalConsiderationItems);
+        _orderParameters.consideration[0].recipient = payable(
+            address(uint160(MAX_INT))
+        );
+        // Get the calldata that will be passed into fulfillOrder.
+        bytes memory fulfillOrderCalldata = abi.encodeWithSelector(
```
stuckinaboot, line 697: Here's the calldata

test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +512,204 @@ contract FulfillOrderTest is BaseOrderTest {
...
+                        orderParamsOffset,
+                        add(considerationItemsOffsetValue, 0x40)
+                    )
+                )
+            )
+
+            // let absoluteLengthOffset := add(orderCalldata, relativeLengthOffset)
+            // length := mload(considerationItemsOffsetValue)
+        }
+        emit log_named_bytes32("output0", output0);
```
stuckinaboot, line 606: Emitting events

test/foundry/FulfillBasicOrderTest.t.sol
```diff
@@ -155,110 +153,32 @@ contract FulfillBasicOrderTest is BaseOrderTest, LowLevelHelpers {
...
-            _revertWithReasonIfOneIsReturned();
-
-            // Otherwise, revert with a generic error message.
-            revert();
-        }
+        _performTestFulfillOrderRevertInvalidArrayLength(
+            consideration,
+            myOrder,
+            fulfillOrderCalldata,
+            0x44,
```
stuckinaboot, line 166: I believe the offset of order parameters in the calldata appears to be at 0x24. However, both here and for FulfillOrderTest.t.sol, I had to use the actual offset of order parameters + 0x20 in order to get (a) the right output in my debug logs (e.g. `emit log_named_bytes32(...)`) (b) get the tests to pass

I'm guessing that's because the length of the calldata is 0x20, and that length field would be stored at the start of the calldata array but would not be printed.

@0age does this sound reasonable?

stuckinaboot: I added this comment, ideally we could clarify this `and 0x20 bytes until the start of order parameters` further but I don't have a good explanation

test/foundry/FulfillOrderTest.t.sol
```diff
@@ -510,6 +510,48 @@ contract FulfillOrderTest is BaseOrderTest {
...
+
+        _performTestFulfillOrderRevertInvalidArrayLength(
+            consideration,
+            _order,
+            fulfillOrderCalldata,
+            // Order parameters starts at 0xa4 relative to the start of the
+            // order calldata because the order calldata starts with 0x20 bytes
+            // for order calldata length, 0x04 bytes for selector, and 0x80
+            // bytes until the start of order parameters.
+            0xa4,
```
stuckinaboot, line 548: Would be good to clarify the 0x80 further but I don't have a good explanation, similar to above https://github.com/ProjectOpenSea/seaport/pull/461/files#r893103417

test/foundry/FulfillBasicOrderTest.t.sol
0age: first there's the function selector (0x00-0x04); then first word (0x04-0x24) is the offset of the OrderParameters struct; next word (0x24-0x44) is the fulfillerConduitKey; _then_ you've got the data for the order parameters at 0x44 in calldata

stuckinaboot: I'm confused, where does the fulfiller's conduit key get included here?
```
bytes memory fulfillOrderCalldata = abi.encodeWithSelector(
            consideration.fulfillBasicOrder.selector,
            _basicOrderParameters
        );
```

stuckinaboot: Resolved offline

PR #464, opened 2022-06-08
Restrict eth offer items

d1ll0n:
Disallow ETH offer items except in matchOrders and matchAdvancedOrders.


d1ll0n:
Working on getting coverage back to 100%

0age:
MagicModulus = 69 ftw 🔥

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -348,6 +356,10 @@ contract ReferenceOrderCombiner is
             }
         }
 
+        if (anyNativeOfferItems && nonMatchFn) {
```
leonardoalt, line 359: Would recommend moving the assignment of `nonMatchFn` here.

contracts/lib/OrderCombiner.sol
```diff
@@ -179,7 +179,29 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
...
+        uint256 invalidNativeOfferItemErrorBuffer;
+        assembly {
+            // Use the second bit of the error buffer to indicate whether we
+            // are in a function that is not matchAdvancedOrders or matchOrders.
+            invalidNativeOfferItemErrorBuffer := shl(
+                1,
+                gt(
+                    // Take the remainder of the selector modulo a magic value.
+                    mod(
+                        shr(NumBitsAfterSelector, calldataload(0)),
```
leonardoalt, line 196: Is this always true? Wouldn't it be safer to just shift by 28 bytes and mask?

contracts/lib/OrderCombiner.sol
```diff
@@ -359,6 +390,13 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
             }
         }
 
+        // If the second bit is set in the error buffer, we are not in a match function.
+        // If the first bit is set, a native offer item was encountered.
+        // If the value is greater than two, both the first and second bits were set.
```
leonardoalt, line 395: This comment is a bit misleading since it's true here because `> 3` is not possible but the statement isn't true in general.

contracts/lib/OrderCombiner.sol
```diff
@@ -179,7 +179,29 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
...
+                    // Take the remainder of the selector modulo a magic value.
+                    mod(
+                        shr(NumBitsAfterSelector, calldataload(0)),
+                        NonMatchSelector_MagicModulus
+                    ),
+                    // Check if the remainder is higher than the greatest remainder
+                    // of the two match selectors modulo the magic value.
+                    NonMatchSelector_MagicRemainder
+                )
+            )
```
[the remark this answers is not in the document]
MrToph: (just a mental note: `0b10` if `fulfill*`, `0b00` otherwise so far after execution)

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -194,6 +194,10 @@ contract ReferenceOrderCombiner is
         // Track the order hash for each order being fulfilled.
         bytes32[] memory orderHashes = new bytes32[](totalOrders);
 
+        // Check if we are in a match function
+        bool nonMatchFn = msg.sig != 0x55944a42 && msg.sig != 0xa8174404;
```
MrToph, line 198: could use `SeaportInterface.matchOrders/matchAdvancedOrders.selector`?

reference/lib/ReferenceOrderCombiner.sol
```diff
@@ -260,6 +264,10 @@ contract ReferenceOrderCombiner is
                 // Retrieve the offer item.
                 OfferItem memory offerItem = offer[j];
 
+                anyNativeOfferItems =
+                    anyNativeOfferItems ||
```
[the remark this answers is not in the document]
MrToph: I don't think it can be removed. It's in a loop, you'd overwrite the result every iteration - only the last iteration would be deciding. You want to check if _any_ of the offer items is native.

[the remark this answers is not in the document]
MrToph: I don't think it can be removed. It's in a loop, you'd overwrite the result every iteration - only the last iteration would be deciding. You want to check if _any_ of the offer items is native.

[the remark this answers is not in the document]
leonardoalt: oh yea my bad, didn't notice the loop in the GH diff

contracts/lib/OrderCombiner.sol
naps62: For reference, this has been fixed in `main`. comment now states:

```
        // If the first bit is set, a native offer item was encountered. If the
        // second bit is set in the error buffer, the current function is not
        // matchOrders or matchAdvancedOrders. If the value is three, both the
        // first and second bits were set; in that case, revert with an error.
```

reference/lib/ReferenceOrderCombiner.sol
leonardoalt: Yea this would be much better

PR #465, opened 2022-06-08
Optimize apply fraction

0age:


reference/lib/ReferenceAmountDeriver.sol
```diff
@@ -35,16 +35,24 @@ contract ReferenceAmountDeriver is AmountDerivationErrors {
...
 
+            // Derive the duration for the order and place it on the stack.
+            uint256 duration = endTime - startTime;
+
+            // Derive time elapsed since the order started & place on stack.
+            uint256 elapsed = block.timestamp - startTime;
+
+            // Derive time remaining until order expires and place on stack.
+            uint256 remaining = duration - elapsed;
+
```
leonardoalt, line 55: This block (and below) are `unchecked` in `AmountDeriver.sol`.

PR #474, opened 2022-06-09
ensure that basic order type is within range

0age:


Thro77le:
Ok. 
`ReferenceAssertions.sol` could also be updated so both are on the same page.

Thro77le:
Ok. 
`ReferenceAssertions.sol` could also be updated so both are on the same page.

PR #475, opened 2022-06-09
Optimize signature verification

d1ll0n:


d1ll0n:
Still getting OOG on tests for InvalidSigner, not sure why

gpersoon:
If `signer == 0`  and  `signature.length` != 27 or 28 then this function could allow the signature.

As a safety precaution: check signer !=0
```solidity 
function _assertValidSignature(address signer, ..., bytes memory signature ) ... {
    address recoveredSigner;
    assembly {
        let signatureLength := mload(signature)
        let lenDiff := sub(ECDSA_MaxLength, signatureLength)
        ...           
        if lt(lenDiff, 2) {    // assume this false
            ...
        }
        ...
        success := eq(signer, recoveredSigner)   // if signer happens to be 0 then success = true
        ...
    }
}
```

gpersoon:
> Still getting OOG on tests for InvalidSigner, not sure why

Maybe because `EIP1271WalletCrossChain.sol` returns a string on revert?
 https://github.com/ProjectOpenSea/seaport/blob/c1711de558a93e039ec93d42c4ea2898c3f55652/contracts/test/EIP1271WalletCrossChain.sol#L131-L136

contracts/lib/SignatureVerification.sol
```diff
@@ -37,119 +35,162 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
...
+                        // Revert with generic invalid signer error message
+                        mstore(0, InvalidSigner_error_signature)
+                        revert(0, InvalidSigner_error_length)
+                    }
+                }
             }
-        } else {
-            // For all other signature lengths, try verification via EIP-1271.
-            // Attempt EIP-1271 static call to signer in case it's a contract.
-            _assertValidEIP1271Signature(signer, digest, signature);
```
Thomasrichard1122, line 94: Valid

Thomasrichard1122 (changes requested):

Chomtana:
> Still getting OOG on tests for InvalidSigner, not sure why

Look like static call even success for EOA case

image

image

0age:
Performed a cleanup pass (including managing stack pressure) and moved to using the existing 1271 test contract as the new one was introducing unnecessary complexity. Could use further review!

Thomasrichard1122:
> _ No description provided. _

Stop and allow it

PR #480, opened 2022-06-09
Add transfer helper

stephankmin:
## Motivation
Add a helper contract to allow bulk transfers of ERC20/ERC721/ERC1155 to a specified recipient.

## Solution
- add `TransferHelper.sol`, `TransferHelperInterface.sol`, and `TransferHelperStructs.sol`
- add Hardhat and Forge tests


stuckinaboot (approved):
Looks great! If possible address the GitHub action lints (about lines too long) but other than that looks good to me

0age (approved):

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                        msg.sender,
+                        recipient,
+                        item.identifier,
+                        item.amount
+                    );
+                }
+            }
+
+            // Call the conduit and execute bulk transfers.
+            ConduitInterface(conduit).execute(conduitTransfers);
```
axic, line 160: There is no validation this returns the magic, so solely `TransferHelper` depends on setup correctly with a valid conduit.

axic: Actually this piece doesn't even use the `conduitController`, but derives an address from from the key. So as long as returndata is returned this is considered successful. Should really check the magic.

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                        );
+                    }
+                }
+            }
+        }
+        // Otherwise, a conduitKey was provided.
+        else {
+            // Derive the conduit address from the deployer, conduit key
+            // and creation code hash.
+            address conduit = address(
```
axic, line 119: I wonder if it would make sense instead having a public function for this on the conduit, though that has a call cost.

contracts/helpers/TransferHelper.sol
axic: and/or should perform the same check the `ConduitController` is doing:  `conduit.codehash == _CONDUIT_RUNTIME_CODE_HASH`.

PR #481, opened 2022-06-09
more gas efficient check for nonexistence of conduit

### LINKED ISSUE #285: ConduitController gas optimisation
In `createConduit` function `conduit.codehash == _CONDUIT_RUNTIME_CODE_HASH` is used to check the existence of conduit. As the conduit contract is not warmed up `EXTCODEHASH` would cost 2500 extra cost. As the storage at `_conduits[conduit]` is anyways updated in future it is better to use `_conduits[conduit].key == bytes32(0)`(or call function `_assertConduitExists`). This will give a gas saving of 2500

P.S. In the end `conduit` contract is warmed up after `CREATE2` but the cost of `CREATE2` is not reduced if the contract is already warmed up by `EXTCODEHASH`

ryanio:
Closes #285

### DIFF
--- a/contracts/conduit/ConduitController.sol
+++ b/contracts/conduit/ConduitController.sol
@@ -86,11 +86,7 @@ contract ConduitController is ConduitControllerInterface {
             )
         );
 
-        // If derived conduit exists, as evidenced by comparing runtime code...
-        if (conduit.codehash == _CONDUIT_RUNTIME_CODE_HASH) {
-            // Revert with an error indicating that the conduit already exists.
-            revert ConduitAlreadyExists(conduit);
-        }
+        _assertConduitDoesNotExist(conduit);
 
         // Deploy the conduit via CREATE2 using the conduit key as the salt.
         new Conduit{ salt: conduitKey }();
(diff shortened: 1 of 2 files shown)


0age:
I think this actually fails for the "zero conduit" which is the one deployed on contract creation.

ryanio:
> I think this actually fails for the "zero conduit" which is the one deployed on contract creation.

hmm well you can't createConduit with the zero key because of the check that the first 20 bytes have to match msg.sender?

but I agree if the method was eventually used somewhere else it wouldn't account for that and could lead to issues.

I could expand the check, which would be skipped in normal circumstances from createConduit:

```
    function _assertConduitDoesNotExist(address conduit) private view {
        // Attempt to retrieve a conduit key for the conduit in question.
        if (
            _conduits[conduit].key != bytes32(0) ||
            (_conduits[conduit].key == bytes32(0) &&
                conduit.codehash == _CONDUIT_RUNTIME_CODE_HASH)
        ) {
            // Revert if conduit was located.
            revert ConduitAlreadyExists(conduit);
        }
    }
```

another question, is the zeroConduit usable? it would fail the current `_assertConduitExists`

0age:
Closing this for the time being as there's very little need to optimize gas on conduit creation; very much on the fence on this one though so could be swayed to reopen

PR #489, opened 2022-06-10
⚡️ AmountDeriver: optimize roundUp

transmissions11:


reference/lib/ReferenceAmountDeriver.sol
```diff
@@ -63,7 +63,6 @@ contract ReferenceAmountDeriver is AmountDerivationErrors {
                 (endAmount * elapsed) +
                 extraCeiling);
 
-            // Division is performed without zero check as it cannot be zero.
```
0age, line 66: Should just replace this with a plain old description of what's happening

Thomasrichard1122: No, it's just a text from her yet

PR #492, opened 2022-06-10
Seaport 1.0 => 1.1

0age:
This PR highlights changes from Seaport 1.0 to Seaport 1.1 and is meant to facilitate review and discussion of those changes.


contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+     * @param conduitKey The key of the conduit through which the bulk transfer
+     *                   should occur.
+     *
+     * @return magicValue A value indicating that the transfers were successful.
+     */
+    function bulkTransfer(
+        TransferHelperItem[] calldata items,
+        address recipient,
+        bytes32 conduitKey
+    ) external override returns (bytes4 magicValue) {
```
gpersoon, line 72: Maybe have reentrancy guards?
Note: with a malicious token, a call to transfer a token could call an arbitrary contract.
This could call the `conduit` or `bulkTransfer`.   (maybe also reentrancy guard in conduit?)

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+     * @param items      The items to transfer.
+     * @param recipient  The address the items should be transferred to.
+     * @param conduitKey The key of the conduit through which the bulk transfer
+     *                   should occur.
+     *
+     * @return magicValue A value indicating that the transfers were successful.
+     */
+    function bulkTransfer(
+        TransferHelperItem[] calldata items,
+        address recipient,
```
gpersoon, line 70: Maybe verify that the `recipient` is able to receive NFT's (in case its a contract)

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                                bytes1(0xff),
+                                address(_CONDUIT_CONTROLLER),
+                                conduitKey,
+                                _CONDUIT_CREATION_CODE_HASH
+                            )
+                        )
+                    )
+                )
+            );
+
```
gpersoon, line 133: maybe check the `conduit` exists / has been deployed.

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                        msg.sender,
+                        recipient,
+                        item.identifier,
+                        item.amount
+                    );
+                }
+            }
+
+            // Call the conduit and execute bulk transfers.
+            ConduitInterface(conduit).execute(conduitTransfers);
```
gpersoon, line 160: maybe check that `return` value is right `magicValue`

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                            recipient,
+                            item.amount
+                        );
+                    } else if (item.itemType == ConduitItemType.ERC721) {
+                        _performERC721Transfer(
+                            item.token,
+                            msg.sender,
+                            recipient,
+                            item.identifier
+                        );
```
gpersoon, line 102: perhaps check `item.amount == 1` like the `Conduit` does

contracts/Seaport.sol
```diff
@@ -1,14 +1,14 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.13;
+pragma solidity >=0.8.13;
```
Thro77le, line 2: For caution and to avoid the unexpected, shouldn't the version be also upper bounded by <0.9.0 or even <=0.8.14

contracts/lib/SignatureVerification.sol
```diff
@@ -36,110 +34,221 @@ contract SignatureVerification is SignatureVerificationErrors, LowLevelHelpers {
...
+                    // Temporarily overwrite the word before the length with
+                    // `digest` to conform to the expected input for ecrecover.
+                    mstore(wordBeforeSignaturePtr, digest)
+
+                    // Attempt to recover the signer for the given signature. Do
+                    // not check the call status as ecrecover will return a null
+                    // address if the signature is invalid.
+                    pop(
+                        staticcall(
+                            Ecrecover_gas, // Gas used by ecrecover precompile.
```
hrkrshnn, line 117: I think this is a bad idea. For one, this creates issues with repricing. Also, this code is more expensive at runtime as well as deploy time. Using `gas()` would be 1 gas cheaper at runtime and also saves 3 bytes at deploy time.

0age: Also I guess it's not like ecrecover can exhaust unnecessary gas anyway?

0age: see #500 @hrkrshnn

contracts/conduit/Conduit.sol
```diff
@@ -167,14 +207,19 @@ contract Conduit is ConduitInterface, TokenTransferrer {
...
+     *      zero-amount items if that constraint is desired.
      *
      * @param item The ERC20/721/1155 item to transfer.
      */
     function _transfer(ConduitTransfer calldata item) internal {
-        // If the item type indicates Ether or a native token...
+        // Determine the transfer method based on the respective item type.
         if (item.itemType == ConduitItemType.ERC20) {
-            // Transfer ERC20 token.
+            // Transfer ERC20 token. Note that item.identifier is ignored and
```
axic, line 219: Is there a particular reason for this?

contracts/conduit/ConduitController.sol
```diff
@@ -198,8 +213,13 @@ contract ConduitController is ConduitControllerInterface {
             revert NewPotentialOwnerIsZeroAddress(conduit);
         }
 
+        // Ensure the new potential owner is not already set.
+        if (newPotentialOwner == _conduits[conduit].potentialOwner) {
```
axic, line 217: This only checks for the same owner being proposed twice, but allows proposing a new one while a proposal is active. Is that the goal?

contracts/helpers/TransferHelper.sol
```diff
@@ -0,0 +1,166 @@
...
+                for (uint256 i = 0; i < totalTransfers; ++i) {
+                    // Retrieve the transfer in question.
+                    TransferHelperItem calldata item = items[i];
+
+                    // Perform a transfer based on the transfer's item type.
+                    // Revert if item being transferred is a native token.
+                    if (item.itemType == ConduitItemType.NATIVE) {
+                        revert InvalidItemType();
+                    } else if (item.itemType == ConduitItemType.ERC20) {
+                        _performERC20Transfer(
```
axic, line 90: This allows `item.identifier` to be anything.

contracts/lib/OrderCombiner.sol
```diff
@@ -175,6 +180,32 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
...
+        uint256 invalidNativeOfferItemErrorBuffer;
+
+        // Use assembly to set the value for the second bit of the error buffer.
+        assembly {
+            // Use the second bit of the error buffer to indicate whether the
+            // current function is not matchAdvancedOrders or matchOrders.
+            invalidNativeOfferItemErrorBuffer := shl(
+                1,
+                gt(
+                    // Take the remainder of the selector modulo a magic value.
```
axic, line 197: :eyes:

Which (two) selectors is this comparing? This seems like an extremely fragile piece of code should new external code paths allowed to reach this function.

axic: It seems to be reachable from `Consideration.matchOrders`, ` Consideration.matchAdvancedOrders`, `Consideration.fulfillAvailableOrders`, `Consideration.fulfillAvailableAdvancedOrders`.

That is 4 selectors this code is comparing. Note that not only new external functions are risky here, but also any parameter change to the above would change the selectors.

axic: Found a prior conversation on the topic: https://github.com/ProjectOpenSea/seaport/pull/464/files#r893417290

contracts/lib/OrderCombiner.sol
```diff
@@ -175,6 +180,32 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
...
+        // {01} == 1 => In a match function, some native offer items: allow.
+        // {10} == 2 => Not in a match function, no native offer items: allow.
+        // {11} == 3 => Not in a match function, some native offer items: THROW.
+        uint256 invalidNativeOfferItemErrorBuffer;
+
+        // Use assembly to set the value for the second bit of the error buffer.
+        assembly {
+            // Use the second bit of the error buffer to indicate whether the
+            // current function is not matchAdvancedOrders or matchOrders.
+            invalidNativeOfferItemErrorBuffer := shl(
```
axic, line 194: Likely no need for assembly on this and it clearly shows how risky it is to the reader:
```solidity
unchecked {
  invalidNativeOfferItemErrorBuffer = ((uint32(msg.selector) % NonMatchSelector_MagicModulus) > NonMatchSelector_MagicRemainder) ? 3 : 0;
}
```

contracts/lib/OrderCombiner.sol
axic: My recommendation would be to pass down a `isMatchOrder` flag.

contracts/conduit/Conduit.sol
0age: Avoiding a redundant check mainly, basically letting channels make the call as to whether they want to do this safety check or not

contracts/conduit/ConduitController.sol
0age: I think that’s fine, the current owner is still in control at that stage

contracts/lib/OrderCombiner.sol
0age: I do like including this as part of the documentation to demonstrate what’s happening here!

contracts/lib/OrderCombiner.sol
0age: Yeah we’ll definitely replace this pattern with one like you suggest should any aspect of the interface change

contracts/conduit/ConduitController.sol
axic: It does feel like a not very useful check though.

contracts/conduit/Conduit.sol
axic: > Avoiding a redundant check mainly,

Where is this checked before hand? The conduit can be used independently, can't it be?

contracts/lib/ConsiderationBase.sol
```diff
@@ -184,7 +193,7 @@ contract ConsiderationBase is ConsiderationEventsAndErrors {
                 "bytes32 zoneHash,",
                 "uint256 salt,",
                 "bytes32 conduitKey,",
-                "uint256 nonce",
+                "uint256 counter",
```
Thro77le, line 196: [Documentation](https://docs.opensea.io/v2.0/reference/seaport-overview) still contains old name  (`nonce` instead of `coutner`)

contracts/helpers/TransferHelper.sol
jameswenzel: What would the benefit of a reentrancy guard be here? In the case that a contract re-enters this function via a callback, won't the `msg.sender` be that contract, in which case, shouldn't it only be able to transfer its _own_ tokens?
I guess worst-case it could be a method of tricking users into paying more gas to transfer a token, but we'd need to decide if eliminating the possibility of that is worth the 7k+ gas overhead on all legitimate transfers – I'd personally say it's not (unless I'm missing something else).

PR #502, opened 2022-06-10
Add Foundry test for failures when spending pre-approved items from null address with invalid signature

jameswenzel:
Test that signature verification works in the case where Seaport has permission to transfer an item at the null address, and fails accordingly.

### DIFF
--- a/test/foundry/FulfillOrderTest.t.sol
+++ b/test/foundry/FulfillOrderTest.t.sol
@@ -107,7 +107,7 @@ contract FulfillOrderTest is BaseOrderTest {
         _;
     }
 
-    function testNoSpendFromNullAddress() public {
+    function testNullAddressSpendReverts() public {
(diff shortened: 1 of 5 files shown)


0age (changes requested):
Can we change the phrasing to make it clear that this is testing for _failure_ when attempting this? Only other thing that we might want to tweak is to use the solmate token already available as a library rather than adding the token as a new contract. Let me know if there’s some detail I’m missing though. (Everything else looks solid!)

jameswenzel:
@0age can definitely make language clearer. Solmate tokens don't have `virtual` modifiers internal methods, and use the generated `public` property getters for ie `isApprovedForAll`, which can't be overridden either. I opted to modify Solmate's 721 rather than add `openzeppelin-contracts` as a Foundry dependency, but we can do that instead if you'd prefer.

PR #510, opened 2022-06-12
Add helper contracts and tests for obtaining an order hash and a digest

iamameme:
## Motivation

As other projects will need to interact with various parts of seaport, add helper contracts for obtaining an order hash and a digest for an order individually.

## Solution

- add `OrderHashHelper.sol`, `DigestHelper.sol`, and `TestOrderHashDigestHelper.sol`
- add Hardhat and Forge tests


0age (changes requested):
Thanks for preparing these helpers! A few things before we can merge:
- We need reference versions of these helpers. So for those, the rule is just no assembly, no unchecked, and make it as readable as possible.
- Be sure to run linters: `yarn lint:fix` and `yarn lint:check`
- Left a few other small nits

iamameme:
> Thanks for preparing these helpers! A few things before we can merge:
> 
> * We need reference versions of these helpers. So for those, the rule is just no assembly, no unchecked, and make it as readable as possible.
> * Be sure to run linters: `yarn lint:fix` and `yarn lint:check`
> * Left a few other small nits

I believe I addressed all these now! Let me know if I missed anything else.

0age (changes requested):
Two remaining things here: first, add function visibility where required by the lint warnings, and second find a way to get code coverage of the missing statement (recommended importing the contract as part of a test contract and hitting it directly as Hardhat doesn’t support support modifying the chainId via a cheat code IIRC)

contracts/helpers/DigestHelper.sol
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.13;
+// prettier-ignore
```
ryanio, line 3: thanks for updating to the new typescript test file! looks great.

two comments here, can you change all the pragma solidity versions to carat `^` instead of `>=` (this has been recently done on main)

Also you won't need this `prettier-ignore` with latest main now having prettier multi-line import support

iamameme: Awesome :D Thanks! Just implemented these changes.

ryanio: If you merge in changes from main it was fixed in a recent PR

ryanio: And looks great! Maybe re-add a newline in this file below the pragma version since it's hugging the first import now.

iamameme: Got that working! Merged main and pushed some linter changes from `lint:fix`. Also added that new line :D

PR #516, opened 2022-06-15
Address TransferHelper comments

stephankmin:


contracts/helpers/TransferHelper.sol
```diff
@@ -3,8 +3,18 @@ pragma solidity >=0.8.7;
 
 import "./TransferHelperStructs.sol";
 
+// prettier-ignore
+import {
+    ERC721TokenReceiver
```
0age, line 8: We should just declare an interface locally rather than importing it from a library, it’s just a single function

contracts/helpers/TransferHelper.sol
```diff
@@ -87,20 +99,59 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
                         _performERC20Transfer(
                             item.token,
                             msg.sender,
                             recipient,
                             item.amount
                         );
                     } else if (item.itemType == ConduitItemType.ERC721) {
+                        // If recipient is a contract, ensure it can receive
+                        // ERC721 tokens.
+                        if (recipient.code.length != 0) {
```
0age, line 117: Might only need to perform this check once, especially as the check occurs out-of-band, since there’s only a single recipient

contracts/helpers/TransferHelper.sol
```diff
@@ -87,20 +99,59 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                            // Check if recipient can receive ERC721 tokens
+                            try
+                                ERC721TokenReceiver(recipient).onERC721Received(
+                                    msg.sender,
```
0age, line 121: Operator should be `address(this)`

contracts/helpers/TransferHelper.sol
```diff
@@ -87,20 +99,59 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                                    selector !=
+                                    ERC721TokenReceiver
+                                        .onERC721Received
+                                        .selector
+                                ) {
+                                    revert InvalidERC721Recipient();
+                                }
+                                // Revert if recipient cannot accept ERC721 tokens.
+                            } catch {
+                                revert InvalidERC721Recipient();
```
0age, line 138: Should bubble up revert reasons if one is present

contracts/helpers/TransferHelper.sol
```diff
@@ -87,20 +99,59 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                            }
+                        }
+                        // Ensure that the amount for an ERC721 transfer is 1.
```
0age, line 141: Newline before this comment

contracts/helpers/TransferHelper.sol
```diff
@@ -131,6 +182,11 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
                 )
             );
 
+            // Check that a conduit exists at the derived address.
+            if (conduit.codehash != _CONDUIT_RUNTIME_CODE_HASH) {
```
0age, line 186: This check is unnecessary; instead, we should retrieve the magic value returned when making the call to the conduit and compare it to the expected value, throwing if it does not equal that value

contracts/test/InvalidERC721Recipient.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.7;
+
+import { ERC721TokenReceiver } from "@rari-capital/solmate/src/tokens/ERC721.sol";
```
0age, line 4: Same here, just use a locally defined interface

0age (changes requested):
Getting close!

stephankmin:
Re-requesting review of changes to contracts while working on HH blocker @0age.

contracts/helpers/TransferHelper.sol
```diff
@@ -131,6 +202,21 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
                 )
             );
 
+            // Create a variable to store the codehash of the conduit.
+            bytes32 codeHash;
+
+            // Retrieve the codehash of the conduit and assign it to codeHash.
+            assembly {
+                codeHash := extcodehash(conduit)
```
0age, line 210: This step is not necessary — instead, we should ensure that the correct magic value is returned when calling the conduit

contracts/helpers/TransferHelper.sol
```diff
@@ -156,8 +242,26 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
                 }
             }
 
-            // Call the conduit and execute bulk transfers.
-            ConduitInterface(conduit).execute(conduitTransfers);
+            // If the external call fails, revert with the conduit's
+            // custom error.
+            try ConduitInterface(conduit).execute(conduitTransfers) {} catch (
```
0age, line 247: So here we’d compare against the magic value rather than just ignoring the return value

contracts/helpers/TransferHelper.sol
ryanio: if there is no contract at the address (i.e. a conduitKey was supplied that doesn't exist) then it won't return a magic value. In fact, it won't do anything at all, except revert with no reason string, with no way we've been able to capture it.

@stephankmin and I found this strange and a similar question [here](https://ethereum.stackexchange.com/questions/128845/why-does-soliditys-try-catch-not-catch-call-to-non-contract-address-revertion) with no answer.

so extcodehash was the cheapest way we're able to identify if it is indeed first a contract account, which uses [400 gas](https://eips.ethereum.org/EIPS/eip-1052#specification). then we can compare the magic value to separate non-conforming conduits.

contracts/helpers/TransferHelper.sol
ryanio: the magic value comparison is now back in, see https://github.com/ProjectOpenSea/seaport/pull/516#discussion_r911557187 for why we found the need for extcodehash before calling the conduit.

contracts/helpers/TransferHelper.sol
ryanio: if there is really no way to capture it (we should keep looking into it) and we think it's worth saving the 400 gas check per transfer then it could just be the defined behavior that would have failed anyway, it just has no message.

contracts/helpers/TransferHelper.sol
```diff
@@ -70,83 +88,111 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
-                    // Create a ConduitTransfer corresponding to each
-                    // TransferHelperItem.
-                    conduitTransfers[i] = ConduitTransfer(
-                        item.itemType,
+                        returns (bytes4 selector) {
+                            // Check if onERC721Received selector is valid.
+                            if (
+                                selector !=
+                                IERC721Receiver.onERC721Received.selector
+                            ) {
```
arr00, line 161: missing branch coverage for statement evaluates to false

contracts/helpers/TransferHelper.sol
```diff
@@ -21,18 +26,28 @@ import {
     TransferHelperInterface
 } from "../interfaces/TransferHelperInterface.sol";
 
+interface IERC721Receiver {
```
0age, line 29: Can we put this in its own file?

contracts/helpers/TransferHelper.sol
```diff
@@ -70,97 +88,242 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                    item.identifier,
+                    item.amount
+                );
+            }
         }
 
-        // Return a magic value indicating that the transfers were performed.
-        magicValue = this.bulkTransfer.selector;
+        // Check if the derived conduit is a contract.
+        if (!_isContract(conduit)) {
```
0age, line 267: This check is unnecessary as a magic value must be returned by the call (and won’t be if there’s no contract) — would lose the InvalidMagicValue custom error and use InvalidConduit in its place.

contracts/helpers/TransferHelper.sol
```diff
@@ -70,97 +88,242 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                // If the external call fails, revert with the conduit key
+                // and conduit address.
+                revert InvalidMagicValue(conduitKey, conduit);
+            }
+        } catch (bytes memory data) {
+            // Catch reverts from the external call to the conduit and
+            // "bubble up" the conduit's revert reason if present.
+            if (data.length != 0) {
+                assembly {
+                    returndatacopy(0, 0, returndatasize())
```
0age, line 291: If we’re being thorough here, we would want to protect against gas exhaustion on loading giant revert reasons into memory; could just throw a generic revert reason if data.length > 256 or something

contracts/helpers/TransferHelper.sol
```diff
@@ -70,97 +88,242 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+            } else {
+                // If no revert reason is present, revert with a generic error
+                // with the conduit key and conduit address.
+                revert ConduitErrorGenericRevert(conduitKey, conduit);
+            }
+        } catch Error(string memory reason) {
+            // Catch reverts with a provided reason string and
+            // revert with the reason, conduit key and conduit address.
+            revert ConduitErrorString(reason, conduitKey, conduit);
+        } catch Panic(uint256 errorCode) {
```
0age, line 303: Is this one necessary?

contracts/helpers/TransferHelper.sol
```diff
@@ -70,97 +88,242 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+        }
+    }
+
+    /**
+     * @notice Check if the passed-in address refers to a contract by comparing
+     *         its codeHash to the default codeHash for non-contract addresses.
+     *
+     * @param account The account to be checked.
+     */
+    function _isContract(address account) internal view returns (bool) {
```
0age, line 316: Can lose this if modifying as suggested above

0age (changes requested):

contracts/helpers/TransferHelper.sol
stephankmin: We’re also using `_isContract` in `_performTransfersWithoutConduit` to check if the recipient is a contract before calling `onERC721Received`. Should we replace with an alternate check or just call `onERC721Received` on all recipients?

ryanio: it looks like `_performERC721Transfer` and `_performERC1155Transfer` already do an extcodesize check:

https://github.com/ProjectOpenSea/seaport/blob/f62c14fe0a94fcb0d26ffdcee48238be63a85b4c/contracts/lib/TokenTransferrer.sol#L248-L253
if these contracts are not already loaded in memory, it would be cheaper to check with extcodehash rather than extcodesize (700 vs 300 gas)

Screen Shot 2022-07-07 at 10 09 19 AM

thus the instances above could also be optimized

contracts/helpers/TransferHelper.sol
```diff
@@ -68,97 +80,223 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                                IERC721Receiver.onERC721Received.selector
+                            ) {
+                                // Revert if recipient cannot accept
+                                // ERC721 tokens.
+                                revert InvalidERC721Recipient();
+                            }
+                        } catch (bytes memory data) {
+                            // "Bubble up" recipient's revert reason
+                            // if present.
+                            if (data.length != 0) {
```
arr00, line 163: Why not limit revert data to something reasonable like conduit return data?

contracts/helpers/TransferHelper.sol
```diff
@@ -68,97 +80,223 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+                } else if (item.itemType == ConduitItemType.ERC721) {
+                    // If recipient is a contract, ensure it can receive
+                    // ERC721 tokens.
+                    if (recipientIsContract) {
+                        // Check if recipient can receive ERC721 tokens.
+                        try
+                            IERC721Receiver(recipient).onERC721Received(
+                                address(this),
+                                msg.sender,
+                                item.identifier,
```
arr00, line 147: Disagree with the premise that we can call this from the transfer helper and get accurate results. Not a view function by interface which means we should accommodate state changes--likely would be a permissioned function. I'd advocate for either making this wholly "unsafe" transfers, or having a flag which allows for real safe transfers

Chomtana: Why not just simply using safeTransferFrom that already perform these checks and call onERC721Received?

arr00: The idea behind this optimization is that it would reduce gas costs.

0age (changes requested):
New batch stuff is good, two notes:
- recommend reusing the same logic across both using a shared internal function
- would advocate against overloading the internal function, makes it less clear what’s being called

contracts/helpers/TransferHelper.sol
```diff
@@ -68,97 +84,428 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+        } else {
+            // Otherwise, a conduitKey was provided.
+            _performTransfersWithConduit(items, recipient, conduitKey);
+        }
 
+        // Return a magic value indicating that the transfers were performed.
+        magicValue = this.bulkTransfer.selector;
+    }
+
+    function bulkTransferToMultipleRecipients(
```
0age, line 99: Can we add natspec here?

contracts/helpers/TransferHelper.sol
```diff
@@ -19,18 +25,25 @@ import {
     TransferHelperInterface
 } from "../interfaces/TransferHelperInterface.sol";
 
+import { TransferHelperErrors } from "../interfaces/TransferHelperErrors.sol";
+
 /**
  * @title TransferHelper
- * @author stuckinaboot, stephankmin
+ * @author stephankmin, stuckinaboot, ryanio
```
stuckinaboot, line 32: I'll let it slide... jk great work

contracts/helpers/TransferHelper.sol
```diff
@@ -42,123 +55,344 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+
+    /**
+     * @notice An internal function that reverts if the passed-in recipient
+     *         is the zero address.
+     *
+     * @param recipient The recipient on which to perform the check.
+     */
+    function _checkRecipientIsNotZeroAddress(address recipient) internal pure {
+        // Revert if the recipient is the zero address.
+        if (recipient == address(0x0)) {
```
stuckinaboot, line 394: Is `address(0)` and `address(0x0)` the same? Is `address(0x0)` more gas optimized?

stuckinaboot (commented):
LGTM, great job

contracts/helpers/TransferHelper.sol
ryanio: I believe they are the same

contracts/helpers/TransferHelper.sol
```diff
@@ -42,123 +55,344 @@ contract TransferHelper is TransferHelperInterface, TokenTransferrer {
...
+
+                    // If the item is an ERC721 token and
+                    // callERC721Receiver is true...
+                    if (item.itemType == ConduitItemType.ERC721) {
+                        // Create a boolean indicating whether
+                        // validateERC721Receiver is true and recipient is
+                        // a contract.
+                        bool callERC721Receiver = transfer
+                            .validateERC721Receiver &&
+                            transfer.recipient.code.length != 0;
```
0age, line 291: This code size check should probably be performed once, outside the loop, and reused here; it’ll increase overhead in cases where there aren’t actually any 721 transfers but will cut down quite a bit in cases where there are a bunch of them

0age: And actually can be skipped when the flag isn’t set anyway!

0age (approved):
⤴️

ryanio (approved):
lgtm, great work!

PR #525, opened 2022-06-18
Updates reference implementation to match #474

naps62:
#474 Introduced a change in both function naming and functionality that apparently wasn't reflected in the reference implementation.

This PR is an attempt to bring that back in sync

Initially reported by @Thro77le on the previous PR

### DIFF
--- a/reference/lib/ReferenceAssertions.sol
+++ b/reference/lib/ReferenceAssertions.sol
@@ -95,23 +95,27 @@ contract ReferenceAssertions is
 
     /**
      * @dev Internal pure function to validate calldata offsets for dynamic
-     *      types in BasicOrderParameters. This ensures that functions using the
-     *      calldata object normally will be using the same data as optimized
-     *      functions. Note that no parameters are supplied as all basic order
-     *      functions use the same calldata encoding.
+     *      types in BasicOrderParameters and other parameters. This ensures
+     *      that functions using the calldata object normally will be using the
+     *      same data as the assembly functions and that values that are bound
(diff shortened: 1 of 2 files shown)


reference/lib/ReferenceAssertions.sol
```diff
@@ -95,23 +95,27 @@ contract ReferenceAssertions is
...
          * 3. Signature offset == 0x240 + (recipients.length * 0x40)
+         * 4. BasicOrderType between 0 and 23 (i.e. < 24)
          */
         // Declare a boolean designating basic order parameter offset validity.
         bool validOffsets = (abi.decode(msg.data[4:36], (uint256)) == 32 &&
             abi.decode(msg.data[548:580], (uint256)) == 576 &&
             abi.decode(msg.data[580:612], (uint256)) ==
-            608 + 64 * abi.decode(msg.data[612:644], (uint256)));
+            608 + 64 * abi.decode(msg.data[612:644], (uint256))) &&
+            abi.decode(msg.data[292:324], (uint256)) < 24;
```
naps62, line 118: sidenote: this being a reference implementation, couldn't this entire function be rewritten in a much more readable way?

0age: 100% — should totally be rewritten to be clearer while accomplishing the same objective!

0age:
Going to merge this in to get to compatibility with optimized contracts, but agree that a refactor of the reference to make the intent clearer is a good idea!

PR #534, opened 2022-06-25
Split `test/index.js` into separate typescript test files

ryanio:
This PR:

* Splits `test/index.js` file into separate files and upgrades them to typescript
* Adds eslint to `lint:check` and `lint:fix`
* Adds eslint `import/order` rule for more consistent import ordering


ryanio:
one note, I renamed `rc-1.1` for the reference contract version to `1.1-reference` to reduce confusion with `rc` typically meaning `release candidate` when used as a version pre/postfix.

test/utils/fixtures/index.ts
```diff
@@ -722,7 +725,7 @@ export const seaportFixture = async (owner: Wallet) => {
         if (consideration.itemType === 1) {
           // ERC20
           // search for transfer
-          const transferLogs = (tokenEvents || [])
+          const transferLogs = (tokenEvents ?? [])
```
stuckinaboot, line 728: Minor: I think you should share with contributors the difference between `||` and `??` as ppl seemed to often use `||` throughout the test files in places where `??` was likely intended (if they knew about it) for "using right side only if left side is null or undefined"... and I have a feeling ppl will keep doing this is if it's not brought up haha

test/utils/helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+    .slice(0, 10);
+
+export const simulateMatchOrders = async (
+  marketplaceContract: ethers.Contract,
+  orders: Order[],
+  fulfillments: Fulfillment[],
+  caller: ethers.Wallet,
+  value: ethers.BigNumberish
+) => {
+  return marketplaceContract
```
stuckinaboot, line 28: nit: can remove `=> {
return ...
}` and just do `=> marketplaceContract...`

(just a minor style thing, your call here)

test/revert.spec.ts
```diff
@@ -0,0 +1,5907 @@
...
+        consideration,
+        0 // FULL_OPEN
+      );
+
+      await expect(
+        marketplaceContract
+          .connect(buyer)
+          .fulfillAvailableOrders(
+            [order],
+            [[[0, 0]]] as any,
```
stuckinaboot, line 5819: nit: is there a more specific type we can use? Does `fulfillAvailableOrders` have typed args?

test/revert.spec.ts
```diff
@@ -0,0 +1,5907 @@
...
+        0 // FULL_OPEN
+      );
+
+      await expect(
+        marketplaceContract
+          .connect(buyer)
+          .fulfillAvailableAdvancedOrders(
+            [order],
+            [],
+            [[[0, 0]]] as any,
```
stuckinaboot, line 5843: Same question as here https://github.com/ProjectOpenSea/seaport/pull/534/files#r907492307

stuckinaboot (approved):
Great job! This will be enourmously helpful going forward and help developers enjoy advancing Seaport tests

test/advanced.spec.ts
```diff
@@ -0,0 +1,4205 @@
...
+        return receipt;
+      });
+
+      orderStatus = await marketplaceContract.getOrderStatus(orderHash);
+
+      expect({ ...orderStatus }).to.deep.equal(
+        buildOrderStatus(true, false, toBN(1), prime2)
+      );
+
+      order.numerator = prime1 as any;
```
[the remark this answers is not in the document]
ryanio: Yes numerator/denominator is number but the issue here is in this test specifically is it is an overflow test so converting to number got a BigNumber overflow error, so passing in as a BigNumber (overriding the type warning with `any`) allowed the overflow error to come from the contract level.

test/advanced.spec.ts
```diff
@@ -0,0 +1,4205 @@
...
+      method: "hardhat_reset",
+    });
+  });
+
+  before(async () => {
+    owner = new ethers.Wallet(randomHex(32), provider);
+
+    await faucet(owner.address, provider);
+
+    ({
```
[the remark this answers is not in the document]
ryanio: It's a good idea, I was thinking about that as I was going but was unsure if it was set up the way it was for a reason. I will give it a try to convert it and it should really simplify the setup as you mention :)

test/utils/fixtures/index.ts
ryanio: Hehe yeah the nullish coalescing operator is really great. Where would you suggest I add this communication? I could add this lint rule actually to help enforce: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-nullish-coalescing.md

0age (approved):
Great work here! Merge at wil

test/advanced.spec.ts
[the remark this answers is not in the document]
ryanio: i'll add a comment to help clarify here

test/revert.spec.ts
ryanio: it does, but it's expecting this generated type from `typechain-types`:
```typescript
export type FulfillmentComponentStruct = {
  orderIndex: BigNumberish;
  itemIndex: BigNumberish;
};
```
any thoughts how to handle so I can remove the `as any` ?

PR #614, opened 2022-08-10
Apply suggested changes from Spearbit

0age:
Main change is to remove "direct" / no-conduit bulk transfer support, as it adds unnecessary surface area — a standalone contract that implements direct transfers should be deployed if and when that functionality is desired (unlike Seaport, there's no composability benefit to colocation of those transfer methods on the same contract).

### DIFF
--- a/contracts/helpers/TransferHelper.sol
+++ b/contracts/helpers/TransferHelper.sol
@@ -272,7 +272,8 @@ contract TransferHelper is TransferHelperInterface, TransferHelperErrors {
      *         to provide an additional layer of assurance that the recipient
      *         can receive the tokens — any hooks or post-transfer checks will
      *         fail and the caller will be the transfer helper rather than the
-     *         ERC721 contract.
+     *         ERC721 contract. Note that the conduit is set as the operator, as
+     *         it will be the caller once the transfer is performed.
      *
      * @param conduit   The conduit to provide as the operator when calling
      *                  onERC721Received.
--- a/contracts/helpers/TransferHelper.sol
+++ b/contracts/helpers/TransferHelper.sol
@@ -5,12 +5,6 @@ import { IERC721Receiver } from "../interfaces/IERC721Receiver.sol";
 
 import "./TransferHelperStructs.sol";
 
-import { TokenTransferrer } from "../lib/TokenTransferrer.sol";
-
-import {
-    TokenTransferrerErrors
-} from "../interfaces/TokenTransferrerErrors.sol";
-
 import { ConduitInterface } from "../interfaces/ConduitInterface.sol";
 
 import {
@@ -33,11 +27,7 @@ import { TransferHelperErrors } from "../interfaces/TransferHelperErrors.sol";
  * @notice TransferHelper is a utility contract for transferring
  *         ERC20/ERC721/ERC1155 items in bulk to specific recipients.
  */
-contract TransferHelper is
-    TransferHelperInterface,
-    TransferHelperErrors,
-    TokenTransferrer
-{
+contract TransferHelper is TransferHelperInterface, TransferHelperErrors {
     // Allow for interaction with the conduit controller.
     ConduitControllerInterface internal immutable _CONDUIT_CONTROLLER;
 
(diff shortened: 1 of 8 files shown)


contracts/helpers/TransferHelper.sol
```diff
@@ -324,36 +215,79 @@ contract TransferHelper is
...
+                customErrorSelector := and(
+                    mload(add(data, 0x20)), // Data begins after length offset.
+                    customErrorSelector
+                )
+            }
+
+            // Pass through the custom error in question if the revert data is
+            // the correct length and matches an expected custom error selector.
+            if (
+                data.length == 4 &&
```
hrkrshnn, line 251: Nit: `data.length >= 4` is more appropriate here. Having extra data in ABI standard is still valid. But fine keeping it to stay consistent with Seaport's use of stricter ABI standard.

contracts/helpers/TransferHelper.sol
```diff
@@ -324,36 +215,79 @@ contract TransferHelper is
...
+
+            // Set initial value of first four bytes of revert data to the mask.
+            bytes4 customErrorSelector = bytes4(0xffffffff);
+
+            // Utilize assembly to read first four bytes (if present) directly.
+            assembly {
+                // Combine original mask with first four bytes of revert data.
+                customErrorSelector := and(
+                    mload(add(data, 0x20)), // Data begins after length offset.
+                    customErrorSelector
```
hrkrshnn, line 244: Note: if `data` is an empty byte array, this could be reading junk value. Looked at the code generated: in the empty case, `data` will be `0x60`--the zero region in memory; So this would actually be reading the 0x80 or the very first free memory pointer location.

However, I don't really see this being problematic. In the worst case, the data at memory location `add(data, 0x20)` is the same as `InvalidItemType.selector` or `InvalidERC721TransferAmount.selector`, and then this would revert with that instead of the correct error `ConduitErrorRevertBytes`, i.e., the worst case is reverting with the wrong custom error.

contracts/helpers/TransferHelper.sol
```diff
@@ -324,36 +215,79 @@ contract TransferHelper is
...
 
     /**
      * @notice An internal function to check if a recipient address implements
-     *         onERC721Received for a given tokenId.
+     *         onERC721Received for a given tokenId. Note that this check does
+     *         not adhere to the safe transfer specification and is only meant
+     *         to provide an additional layer of assurance that the recipient
+     *         can receive the tokens — any hooks or post-transfer checks will
+     *         fail and the caller will be the transfer helper rather than the
+     *         ERC721 contract.
```
hrkrshnn, line 275:
```suggestion
     *         conduit.
```
Perhaps add something along the lines of "operator is faked to be the conduit".

https://eips.ethereum.org/EIPS/eip-721
> /// @param _operator The address which called `safeTransferFrom` function

hrkrshnn: Ah no, the caller is the 721 contract, my bad.

contracts/helpers/TransferHelper.sol
```diff
@@ -81,113 +71,18 @@ contract TransferHelper is
...
-                // Retrieve the items of the transfer in question.
-                TransferHelperItem[] calldata transferItems = transfer.items;
-
-                // Ensure recipient is not the zero address.
-                _checkRecipientIsNotZeroAddress(transfer.recipient);
-
-                // Create a boolean indicating whether validateERC721Receiver
-                // is true and recipient is a contract.
-                bool callERC721Receiver = transfer.validateERC721Receiver &&
-                    transfer.recipient.code.length != 0;
```
arr00, line 126: Note that checking code was removed. If you try to check the receiver on an EOA, it won't return magic value.

contracts/helpers/TransferHelper.sol
0age: Yeah also data.length won’t be 4 in that case so the value read will just be ignored

contracts/helpers/TransferHelper.sol
0age: Yeah also don’t want to risk false positives on errors with the same signature but with arguments

contracts/helpers/TransferHelper.sol
0age: done

PR #646, opened 2022-09-07
⛽️ added gas profiling for the reference contracts

PatrickAlphaC:
## Motivation

Part of the rationale behind the reference contracts is to show how much more gas efficient the yul optimizations are. I've added a gas profiling command for the reference contracts to make it easier for people to run the comparisons.


hardhat-reference.config.ts
```diff
@@ -28,15 +28,47 @@ const config: HardhatUserConfig = {
   solidity: {
     compilers: [
       {
-        version: "0.8.7",
+        version: "0.8.14",
```
0age, line 31: This version was selected for the reference contracts deliberately as it's a "stable" compiler version (in contrast to more recent versions which lean into the yul IR pipeline)

PatrickAlphaC: Makes sense.

PR #670, opened 2022-10-06
use push4 for isValidOrder selector

0age:


0age:
seems to be increasing bytecode size by quite a bit... closing for now

PR #685, opened 2022-11-10
change SLOAD to MLOAD

MollyHe5:
## Motivation
Reading a state variable costs more gas than reading a local variable. In the following example, function assignOperator0 costs 89 more gases than assignOperator.
```
contract LibDemo {
    address public operator;

    event OperatorUpdated(address newOperator);

    function assignOperator0(address operatorToAssign) external {
         // Set the given address as the new operator.
        operator = operatorToAssign;

        // Emit an event indicating the operator has been updated.
        emit OperatorUpdated(operator);
    }

    function assignOperator(address operatorToAssign) external {
         // Set the given address as the new operator.
        operator = operatorToAssign;

        // Emit an event indicating the operator has been updated.
        emit OperatorUpdated(operatorToAssign);
    }
}
```

## Solution
If a state variable is read immediately after assignment, then replacing this read directly with the previously assigned local variable will change one SLOAD to one MLOAD to save some gases.

### DIFF
--- a/contracts/zones/PausableZone.sol
+++ b/contracts/zones/PausableZone.sol
@@ -127,7 +127,7 @@ contract PausableZone is
         operator = operatorToAssign;
 
         // Emit an event indicating the operator has been updated.
-        emit OperatorUpdated(operator);
+        emit OperatorUpdated(operatorToAssign);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


songlh:
@0age Can you take a look at this pull request?

PR #705, opened 2022-12-01
1.2 post execution testing - interface change

jameswenzel:


androolloyd:
for contract types would the assumption be that the same data is being provided?

PR #708, opened 2022-12-02
remove offerer bypass on restricted orders; update forge test

jameswenzel:

### DIFF
--- a/contracts/lib/ZoneInteraction.sol
+++ b/contracts/lib/ZoneInteraction.sol
@@ -38,15 +38,10 @@ contract ZoneInteraction is ZoneInteractionErrors, LowLevelHelpers {
         address offerer,
         address zone
     ) internal view {
-        // Order type 2-3 require zone or offerer be caller or zone to approve.
-        bool isRestricted;
-        assembly {
-            isRestricted := or(eq(orderType, 2), eq(orderType, 3))
-        }
+        // Order type 2-3 require zone to be caller or zone to approve.
         if (
-            isRestricted &&
-            !_unmaskedAddressComparison(msg.sender, zone) &&
-            !_unmaskedAddressComparison(msg.sender, offerer)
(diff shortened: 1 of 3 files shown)


reference/lib/ReferenceZoneInteraction.sol
```diff
@@ -41,12 +41,8 @@ contract ReferenceZoneInteraction is ZoneInteractionErrors {
...
         address zone
     ) internal view {
-        // Order type 2-3 require zone or offerer be caller or zone to approve.
-        if (
-            uint256(orderType) > 1 &&
-            msg.sender != zone &&
-            msg.sender != offerer
-        ) {
+        // Order type 2-3 require zone to be caller or zone to approve.
+        if (uint256(orderType) > 1 && msg.sender != zone) {
```
0age, line 45: This should check for the order types explicitly as it'll trip on 4 (contract)

contracts/lib/ZoneInteraction.sol
```diff
@@ -147,11 +146,10 @@ contract ZoneInteraction is ZoneInteractionErrors, LowLevelHelpers {
         address offerer,
         address zone
     ) internal view {
-        // Order type 2-3 require zone or offerer be caller or zone to approve.
+        // Order type 2-3 require zone to be caller or zone to approve.
         if (
             uint256(orderType) > 1 &&
```
0age, line 151: check for 2-3 explicitly

reference/lib/ReferenceZoneInteraction.sol
```diff
@@ -94,12 +90,8 @@ contract ReferenceZoneInteraction is ZoneInteractionErrors {
...
```
0age, line 94: check for 2-3 explictly (no 4)

0age (changes requested):

0age:
i think this is old now?

PR #713, opened 2022-12-05
expand OrderValidated event + trim contract size to make it fit

0age:


St4rgarden:
I am wondering if decoding the OrderParameters will actually be easy without knowing the length of the OfferItem array.  We can get the original consideration array length from the end of the OrderParameters log - but I am wondering if knowing the length of the OfferItem array might make decoding the event easier.

PR #722, opened 2022-12-06
Use Hardhat Chai Matchers instead of hardhat-waffle

fvictorio:
[Hardhat Chai Matchers](https://hardhat.org/hardhat-chai-matchers/docs/overview) works better than the hardhat-waffle plugin, especially when it comes to asserting revert reasons.

This PR replaces hardhat-waffle with Hardhat Chai Matchers, and adapts the tests in a single suite.

I also changed the version of Hardhat used, from @0age's fork to a beta version that adds better support for the IR pipeline. We are going to include those changes in the latest stable version soon.

I didn't push the changes to the lock files, in part because I don't understand why there is both a `package-lock.json` and a `yarn.lock`, but also because I think it's better if only team members make changes to the lock files.


test/revert.spec.ts
```diff
@@ -176,7 +177,7 @@ describe(`Reverts (Seaport v${VERSION})`, function () {
               value,
             }
           )
-      ).to.be.revertedWith("BadFraction");
+      ).to.be.revertedWithCustomError(marketplaceContract, "BadFraction");
```
fvictorio, line 180: `revertedWithCustomError` needs a reference to the contract that defines the custom error

test/revert.spec.ts
```diff
@@ -3146,9 +3310,7 @@ describe(`Reverts (Seaport v${VERSION})`, function () {
         marketplaceContract
           .connect(owner)
           .matchAdvancedOrders([order, order2], [], fulfillments)
-      ).to.be.revertedWith(
-        "panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)"
-      );
+      ).to.be.revertedWithPanic(PANIC_CODES.ARITHMETIC_UNDER_OR_OVERFLOW);
```
fvictorio, line 3313: This is another new method that is available in Hardhat Chai Matchers and not in waffle.

test/revert.spec.ts
```diff
@@ -4489,7 +4687,7 @@ describe(`Reverts (Seaport v${VERSION})`, function () {
               value,
             }
           )
-      ).to.be.revertedWith(`NOT_AUTHORIZED`);
+      ).to.be.revertedWith("NOT_AUTHORIZED");
```
fvictorio, line 4690: `.revertedWith` now checks that the return data is an encoded `Error(string)`, and performs an exact match with the reason string in the revert.

PR #738, opened 2022-12-07
add some inline comments

DJViau:
## Motivation

Adds inline comments, mostly to the reference implementations, for the sake of clarity.

## Solution

I may have relied on GPT3 and GitHub Copilot a bit too heavily in my effort to move quickly and make up for my lack of understnading, please make sure that these make sense.

### DIFF
--- a/contracts/lib/OrderValidator.sol
+++ b/contracts/lib/OrderValidator.sol
@@ -128,13 +128,18 @@ contract OrderValidator is Executor, ZoneInteraction {
             return (bytes32(0), 0, 0);
         }
 
-        if (orderParameters.orderType == OrderType.CONTRACT) {
-            return
-                _getGeneratedOrder(
-                    orderParameters,
-                    advancedOrder.extraData,
-                    revertOnInvalid
-                );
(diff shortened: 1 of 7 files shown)


contracts/lib/OrderValidator.sol
```diff
@@ -369,7 +374,8 @@ contract OrderValidator is Executor, ZoneInteraction {
                 orderParameters.offer = extendedOffer;
             }
 
-            // Loop through offer and ensure at least as much on returned offer
+            // Loop through each offer and ensure at least as much on returned
```
BCLeFevre, line 377: 🤔 
This sounds weird.

DJViau: It was written by 0 and I was scared to materially change it, so I just tried to make it a wee bit clearer 😅

jameswenzel: "Loop through each new offer and ensure the new amounts are at least as much as the respective original amounts"?

DJViau: Sounds good.  Updated

PR #744, opened 2022-12-08
add comments to test pool contracts

DJViau:
## Motivation

I kinda just want to get more familiar with the code at this point, but this can't hurt, right?

## Solution

Added a bunch of comments to the test pool contracts.


jameswenzel (approved):
Looks great, thanks Dan!

PR #754, opened 2022-12-09
Add automated gas report tracking

d1ll0n:
Updates hardhat-gas-reporter config to produce an output file with the gas report, and adds tasks to process and compare these reports over time.


d1ll0n:
wrong base branch

PR #791, opened 2022-12-17
Use loadFixture in Hardhat tests

fvictorio:
This PR adds `@nomicfoundation/hardhat-network-helpers` and uses its [`loadFixture`](https://hardhat.org/hardhat-network-helpers/docs/reference#loadfixture()) functionality in all `beforeEach`s.

On my machine, this makes `test:ref` go from ~106s to ~86s.


test/revert.spec.ts
```diff
@@ -5800,7 +5810,7 @@ describe(`Reverts (Seaport v${VERSION})`, function () {
     let seller: Wallet;
     let buyer: Wallet;
 
-    before(async () => {
+    beforeEach(async () => {
```
fvictorio, line 5813: This is necessary because `loadFixture` uses snapshots to revert the network to a given state. This combines with the fact that mocha's `before` hooks are executed before any `beforeEach` (even the ones that are at an upper level). So here what happens is:

- This `before` executes first
- The `beforeEach` reverts to the fixture's snapshots and undoes anything done here

Using a `beforeEach` fixes that. It's a bit more wasteful, but I checked and it's still worth it, performance-wise.

0age:
This is great! I adjusted the target branch to 1.2 as that's the main branch we're working off of now; think you could resolve the conflicts there?

PR #793, opened 2022-12-19
ZoneInteraction comment tweaks

0age:


contracts/lib/ZoneInteraction.sol
```diff
@@ -44,11 +44,11 @@ contract ZoneInteraction is
...
 {
     /**
-     * @dev Internal view function to determine if an order has a restricted
-     *      order type and, if so, to ensure that either the offerer or the zone
-     *      are the fulfiller or that a staticcall to `isValidOrder` on the zone
-     *      returns a magic value indicating that the order is currently valid.
-     *      Note that contract orders are not accessible via basic fulfillments.
+     * @dev Internal function to determine if an order has a restricted order
+     *      type and, if so, to ensure that either the offerer or the zone are
+     *      the fulfiller or that a call to `validateOrder` on the zone returns
```
Saw-mon-and-Natalie, line 49: It does not check if the `offerer` is the `fulfiller` for restricted orders.

PR #795, opened 2022-12-20
Rework the signature verification tests

DJViau:
## Motivation

We want to get to 100% coverage on signature verification.

## Solution

We do a little testing.


jameswenzel:
Probably should remove the empty `TokenTransferrer.t.sol` file - but otherwise looks great!

PR #802, opened 2022-12-22
WIP - Dan/2022/12/test token transferrer

DJViau:
## Motivation

We want to add more tests.

## Solution

I'm adding tests.

Before:
1 2r

After:
my-branch
my-branch-

Some of the paths in TokenTransferrer.sol are going to be a pain to hit.  I might keep playing around, but since I've only got roughly a half day left before the holidays, I wanted to get the ball rolling on getting this merged.

As always, I'd welcome feedback at any level: approach, structure, style, etc.


test/foundry/TokenTransferrer.t.sol
```diff
@@ -0,0 +1,274 @@
...
+            ConduitItemType.ERC20,
+            ConduitItemType.ERC721,
+            ConduitItemType.ERC1155
+        ];
+        ConduitItemType itemType;
+
+        address noCodeTokenAddress;
+        noCodeTokenAddress = address(0xabc);
+
+        uint256 alicePk = 0xa11ce;
```
jameswenzel, line 52: you can use the `forge-std` helper `makeAddr('some string')` to make various addresses; it will also label them with whatever string you provide in forge stack traces - neat!

test/foundry/TokenTransferrer.t.sol
```diff
@@ -0,0 +1,274 @@
...
+        //         address(tokenRevert),
+        //         address(alice),
+        //         address(bob),
+        //         0,
+        //         1
+        //     )
+        // );
+        // No bubbling up in the reference contract.
+        vm.expectRevert();
+        test(
```
jameswenzel, line 152: Since these are supposed to revert - I think this makes more sense if you just call `this.execute` - should (hopefully) have the same effect as the wrapped `test`, since no state will actually change as a result

test/foundry/TokenTransferrer.t.sol
```diff
@@ -0,0 +1,274 @@
...
+        revertTransfer[0] = ConduitTransfer(
+            ConduitItemType.ERC721,
+            address(nonfungibleTokenRevert),
+            address(alice),
+            address(bob),
+            0,
+            1
+        );
+
+        // No bubbling up in the reference contract.
```
jameswenzel, line 212: if the reason we're explicitly calling `test` on different implementations rather than passing them via context is because they revert differently - we could add a state variable `bool reference` (or similar) to conditionally expect one error or another without duplicating the calls

test/foundry/TokenTransferrer.t.sol
```diff
@@ -0,0 +1,274 @@
...
+        //         1
+        //     )
+        // );
+        // No bubbling up in the reference contract.
+        vm.expectRevert();
+        test(
+            this.execute,
+            Context(referenceConduit, revertTransfer, noCodeBatchTransfer)
+        );
+        vm.expectRevert("Some ERC20 revert message");
```
jameswenzel, line 156: I'm also a little confused why this is working... I feel like the `test` wrapper theoretically should just revert with `DifferentialTestAssertionFailed()` when it reverts, but also I know that's not always the case 🤔  Maybe because it's a string revert?

jameswenzel:
LGTM! Clever use of context to pass the revert reason 🧠

PR #805, opened 2022-12-27
Fix foundry fuzz config

paulrberg:
The `fuzz_runs` and `fuzz_max_global_rejects` have been changed to be sub-properties in a new `fuzz` inline table.


paulrberg:
Closing, since this seems to be fixed on the `1.2` branch.

PR #813, opened 2023-01-03
general cleanup

0age:


contracts/lib/TokenTransferrerConstants.sol
```diff
@@ -119,6 +121,7 @@ uint256 constant TokenTransferGenericFailure_error_token_ptr = 0x20;
 uint256 constant TokenTransferGenericFailure_error_from_ptr = 0x40;
 uint256 constant TokenTransferGenericFailure_error_to_ptr = 0x60;
 uint256 constant TokenTransferGenericFailure_error_identifier_ptr = 0x80;
+uint256 constant TokenTransferGenericFailure_err_identifier_ptr = 0x80;
```
Saw-mon-and-Natalie, line 124: Is `TokenTransferGenericFailure_err_identifier_ptr` redundant?

`TokenTransferGenericFailure_error_identifier_ptr` is already defined with the same value.

The only difference is when `TokenTransferGenericFailure_err_identifier_ptr` is used the `identifier` is `0` for `ERC20` transfers.

PR #834, opened 2023-01-05
1.2 Optimizes calculating the `invalidNativeOfferItemErrorBuffer` by using a bitmask

Saw-mon-and-Natalie:
Simpler formula to mark non-match selectors. Addresses:
- https://github.com/spearbit-audits/review-seaport-1.2/issues/80


0age:
https://github.com/ProjectOpenSea/seaport/pull/839 modifies the `matchAdvancedOrders` selector, which likely breaks both this and the original "magic modulus" math. Guessing this logic will need to be revisited, but that there must still be a byte that works!

0age:
Implemented as part of https://github.com/ProjectOpenSea/seaport/pull/851 (would appreciate review @Saw-mon-and-Natalie!)

PR #842, opened 2023-01-06
Modify fulfillment aggregation functions to handle all fulfillments i…

0age:
…n same loop and skip OOR items


contracts/lib/FulfillmentApplier.sol
```diff
@@ -225,262 +234,185 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
 
                 // Update error buffer: 1 = zero amount, 2 = overflow, 3 = both.
                 errorBuffer := or(
-                  errorBuffer,
-                  or(
-                    shl(1, lt(newAmount, amount)),
-                    iszero(mload(amountPtr))
-                  )
+                    errorBuffer,
+                    or(shl(1, lt(newAmount, amount)), iszero(mload(amountPtr)))
```
Saw-mon-and-Natalie, line 408: might be cheaper to cache `mload(amountPtr)` as it has been also used on line 403.

contracts/lib/FulfillmentApplier.sol
```diff
@@ -528,244 +484,167 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
+                {
+                    // Retrieve amount pointer using consideration item pointer.
+                    let amountPtr := add(
+                        considerationItemPtr,
+                        Common_amount_offset
+                    )
 
-                // Add offer amount to execution amount.
-                let newAmount := add(amount, mload(amountPtr))
+                    // Add offer amount to execution amount.
```
Saw-mon-and-Natalie, line 566: typo, needs to be `consideration amount`.

contracts/lib/FulfillmentApplier.sol
```diff
@@ -225,262 +234,185 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
+                    iszero(lt(itemIndex, mload(offerArrPtr))),
+                    iszero(mload(add(orderPtr, AdvancedOrder_numerator_offset)))
                 ) {
-                    throwInvalidFulfillmentComponentData()
+                    continue
                 }
 
-                // Retrieve offer item pointer using index.
-                offerItemPtr := mload(
+                // Retrieve consideration item pointer using the item index.
```
Saw-mon-and-Natalie, line 297: typo, needs to mention `offer item pointer`.

contracts/lib/FulfillmentApplier.sol
```diff
@@ -528,244 +484,167 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
-                // Update the amount to the new, summed amount.
-                amount := newAmount
+                    // Update the amount to the new, summed amount.
+                    amount := newAmount
 
-                // Zero out amount on original item to indicate it is credited.
-                mstore(amountPtr, 0)
+                    // Zero out amount on original item to indicate it is credited.
+                    mstore(amountPtr, 0)
+                }
```
Saw-mon-and-Natalie, line 583: For offer item aggeregation subroutine this part comes after the `switch` block.

Saw-mon-and-Natalie: It would be great to keep the flow consistent for both functions.

Saw-mon-and-Natalie: It is important to keep this block in this function before the `switch` block, due to the hashing comparison that requires the `amount` slot to be zeroed out.

contracts/lib/FulfillmentApplier.sol
```diff
@@ -225,262 +234,183 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
+                    add(mload(fulfillmentHeadPtr), Fulfillment_itemIndex_offset)
+                )
 
-                // Throw if itemIndex is out of the range of array.
-                if iszero(
-                    lt(itemIndex, mload(offerArrPtr))
+                // If the offer item index is out of range or the numerator
+                // is zero, skip this item.
+                if or(
+                    iszero(lt(itemIndex, mload(offerArrPtr))),
```
Saw-mon-and-Natalie, line 291: Was there a particular reason, we also choose to skip OORs for offer array item index? Orders of `CONTRACT` order type cannot reduce the length of this array.

contracts/lib/FulfillmentApplier.sol
```diff
@@ -225,262 +231,183 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
-                              )
-                          )
+                    // If component index > 0, swap component pointer with pointer
+                    // to first component.
+                    let firstFulfillmentHeadPtr := add(offerComponents, OneWord)
+                    if xor(firstFulfillmentHeadPtr, fulfillmentHeadPtr) {
+                      let firstFulfillmentPtr := mload(firstFulfillmentHeadPtr)
+                      let fulfillmentPtr := mload(fulfillmentHeadPtr)
+                      mstore(firstFulfillmentHeadPtr, fulfillmentPtr)
+                      mstore(fulfillmentHeadPtr, firstFulfillmentPtr)
```
Saw-mon-and-Natalie, line 353: This makes sure that when there is a leftover unmatched aggregated offer item amounts, this amount will be overwritten to (now) the first used offer element that is pointed to by the first offer fulfillment element.

The PR does make a similar change on the consideration aggregation side. So for example, when `Seaport` tries to [store a leftover amount](https://github.com/ProjectOpenSea/seaport/blob/c7c1645e149b4e6840cb7e4f92d687eb5ba8fbcc/contracts/lib/FulfillmentApplier.sol#L105-L117) for consideration if the first element of the consideration fulfillment points to a consideration index that is OOR the transaction would fail (even though, we skipped this item in the aggregation subroutine).

contracts/lib/FulfillmentApplier.sol
0age: had to do this to manage stack pressure on the coverage tests 🙃

contracts/lib/FulfillmentApplier.sol
```diff
@@ -70,6 +70,15 @@ contract FulfillmentApplier is FulfillmentApplicationErrors {
...
         ReceivedItem memory considerationItem = considerationExecution.item;
 
+        // Skip aggregating offer items if no consideration items are available.
+        if (considerationItem.amount == 0) {
+            // Set the offerer and recipient to null address if execution
+            // amount is zero. This will cause the execution item to be skipped.
+            considerationExecution.offerer = address(0);
+            considerationExecution.item.recipient = payable(0);
+            return considerationExecution;
+        }
```
Saw-mon-and-Natalie, line 80: If a contract offerer decides to gift the offer items by removing all the consideration items, the execution will be skipped.

PR #847, opened 2023-01-09
test: criteria based contract orders for the non-wildcard cases

hrkrshnn:
Added a test where a Merkle-proof is *required*.

See: https://github.com/spearbit-audits/review-seaport-1.2/issues/111


hrkrshnn:
There are also some other foundry based tests in https://github.com/ProjectOpenSea/seaport/pull/822/commits/c4da5e56f3e8d360a07eab371224be38a8e34b8d

PR #849, opened 2023-01-09
document contract order collection offer behavior

DJViau:
Adds documentation on a case where contract order behavior differs from normal order behavior.

### DIFF
--- a/reference/lib/ReferenceOrderValidator.sol
+++ b/reference/lib/ReferenceOrderValidator.sol
@@ -268,11 +268,11 @@ contract ReferenceOrderValidator is
     }
 
     /**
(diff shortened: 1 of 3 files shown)


reference/lib/ReferenceOrderValidator.sol
```diff
@@ -267,6 +267,25 @@ contract ReferenceOrderValidator is
         return (orderHash, uint120(numerator), uint120(denominator));
     }
 
+    /**
+     * @dev Internal function to generate a contract order. If you supply a
```
0age, line 271: Can we remove the "you" here and just keep the comments statement-of-fact?

PR #850, opened 2023-01-09
Fix Natspec spacing

Ellahinator:
Addresses:
> The `@` NatSpec fields have an extra space in `Consideration.sol`:
> ```solidity
> * @ <field>
> ```
> The extra space can be removed.

- https://github.com/spearbit-audits/review-seaport-1.2/issues/47


0age:
fixed up in a slightly different fashion!

PR #884, opened 2023-01-17
[docs] Start a cookbook guide

aegas-io:
## Motivation

Hi Seaport team, thanks for the great job you did and continue doing with the protocol! When I first started looking into it, the available docs were describing the big picture and the solution robustness well, but I found it a bit hard to understand how certain things work in practice. Discussions also have questions like "How to do X on Seaport", and there are open issues requesting additional docs.

I started creation of the practical cookbook-like guide. I believe this can be useful for newcomers who want to understand how the protocol works, and it will also encourage more developers to build their marketplaces using Seaport and to invent new sale mechanics.

The cookbook definitely isn't final and can be appended by new sale mechanics.

## Solution

Cookbook docs introduced at `/docs/cookbook/` with:
- ELI5 review of Orders, Order object signing, Order fulfillment
- Guides on Fixed-price sale, Offer-based sale, English auction, Collection offer and Multiple assets trade mechanics

### DIFF
--- a/docs/cookbook/1_Overview_ELI5.md
+++ b/docs/cookbook/1_Overview_ELI5.md
@@ -1,0 +1,62 @@
+# Overview ELI5
+
+## What Order is
+
+Trades within the Seaport protocol are called Orders. Some examples of orders are Fixed-price sales, English or Dutch auction sales, Offer-based sale, and Collection offers.
+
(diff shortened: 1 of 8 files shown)


aegas-io:
Hi Seaport team, it would be nice to get feedback on the guide if you have a chance to take a look. As a few issue creators mentioned it as helpful, I'd continue updating it for the new protocol versions.

AmmarKhalid123:
This PR is very helpful and worth merging into the current repository. ++

PR #935, opened 2023-01-27
address final spearbit requested changes

0age:


contracts/lib/ZoneInteraction.sol
```diff
@@ -118,17 +122,27 @@ contract ZoneInteraction is
             // Set the restricted-order-specific error selector.
             errorSelector = InvalidRestrictedOrder_error_selector;
         } else if (parameters.orderType == OrderType.CONTRACT) {
+            // Set the target to the offerer.
+            target = parameters.offerer;
```
Saw-mon-and-Natalie, line 126: `solc` performs a `push20 0xff...ff` masking here.

contracts/lib/ConsiderationDecoder.sol
```diff
@@ -909,11 +911,15 @@ contract ConsiderationDecoder {
...
-                        // or is not equal to the calculated size.
+                        // or is greater than the calculated size.
                         invalidEncoding := or(
-                            gt(or(offerLength, considerationLength), 0xffff),
-                            xor(totalSize, returndatasize())
+                            gt(
+                                or(offerLength, considerationLength),
+                                generateOrder_maximum_returndatasize
+                            ),
+                            gt(totalSize, returndatasize())
```
Saw-mon-and-Natalie, line 920: here we are allowing extra returned bytes than needed.

contracts/lib/ConsiderationEncoder.sol
```diff
@@ -253,7 +255,8 @@ contract ConsiderationEncoder {
         bytes32 orderHash, // e.g. offerer + contract nonce
```
Saw-mon-and-Natalie, line 255: comment can be more accurate

contracts/lib/ConsiderationDecoder.sol
hrkrshnn: That's consistent with the ABI standard--even though Seaport tries to be stricter at times.

contracts/lib/ZoneInteraction.sol
hrkrshnn: Yeah, this mask is redundant because of the shift later. Would be nice to generalize this and add into the rulelist in solc. Do you want to give it a try @Saw-mon-and-Natalie?

hrkrshnn: But the target value is used below. So maybe it's safe to have the mask (assuming it's there), even though it's used for a `call(...)` later on where the upper bits would be ignored.

PR #968, opened 2023-02-09
add docs for bulk sigs and contract orders

DJViau:
## Motivation

It'd be useful to have an entry point for engineers who want to take advantage of the new big features in 1.2.

## Solution

Add some docs on bulk order creation and contract orders.

EDIT: Easier to review here https://github.com/ProjectOpenSea/seaport/blob/dan/2023/02/add-bulk-sig-contract-order-docs/docs/SeaportDocumentation.md#contract-orders


docs/SeaportDocumentation.md
```diff
@@ -166,6 +164,379 @@ When matching a group of orders via `matchOrders` or `matchAdvancedOrders`, step
...
+        return ("ExampleContractOfferer", schemas);
+    }
+}
+```
+
+## Bulk Order Creation
+
+Seaport v1.2 introduced a bulk order creation feature.  In brief, a buyer or seller can now sign a single bulk order payload that creates multiple orders with one ECDSA signature.  So, instead of signing a dozen single order payloads to create a dozen orders, a user can now create the same dozen orders with a single click in their wallet UI.
+
+Bulk signature payloads of depth 1 (2 orders) to depth 24 (16,777,215 orders) are fully supported as of v1.2.  Just as with single order signing, bulk order payloads will be typed, human-readable EIP 712 data.  Any individual order created in the course of bulk order creation is fulfillable independently.  In other words, one order or multiple orders created in the course of bulk order creation can be included in a fulfillment transaction.
```
DJViau, line 385: 16,777,216*

docs/SeaportDocumentation.md
```diff
@@ -166,6 +164,407 @@ When matching a group of orders via `matchOrders` or `matchAdvancedOrders`, step
...
+            Schema[] memory schemas
+        );
+}
+```
+
+See the [TestContractOfferer.sol](https://github.com/ProjectOpenSea/seaport/blob/main/contracts/test/TestContractOfferer.sol) file in `./contracts/test/` for an example of an MVP Seaport app contract.
+
+### Arguments and Basic Functionality
+
+When Seaport receives a contract order request from an EOA or a [1271](https://eips.ethereum.org/EIPS/eip-1271) contract, it calls the Seaport app contract’s `generateOrder` function, which returns an array of `SpentItem`s and an array of `ReceivedItem`s.  The Seaport app can adjust the response according to its own rules and if its response falls within the acceptable range specified in the original requester's offer (`minimumReceived`) and consideration (`maximumSpent`) parameters, Seaport will execute the orders.  If not, the call will revert.
```
0age, line 222:
```suggestion
When Seaport receives a contract order request from a fulfiller, it calls the Seaport app contract’s `generateOrder` function, which returns an array of `SpentItem`s and an array of `ReceivedItem`s.  The Seaport app can adjust the response according to its own rules and if its response falls within the acceptable range specified in the original requester's offer (`minimumReceived`) and consideration (`maximumSpent`) parameters, Seaport will execute the orders.  If not, the call will revert.
```

docs/SeaportDocumentation.md
```diff
@@ -57,21 +59,17 @@ Orders are fulfilled via one of four methods:
...
+> Note: Calls to Seaport that would fulfill or match a collection of advanced orders can be monitored and where there are unused offer items, it's possible for a third party to claim them. Anyone can monitor the mempool to find calls to `fulfillAvailableOrders`, `fulfillAvailableAdvancedOrders`, `matchOrders`, `matchAdvancedOrders` and calculate if there are any unused offer item amounts.  If there are unused offer item amounts, the third party can create orders with no offer items, but with consideration items mirroring the unused offer items and populate the fulfillment aggregation data to match the unused offer items with the new mirrored consideration items.  This would allow the third party to claim the unused offer items.  A Seaport app or a zone could prevent this, but by default, it's possible.
```
0age, line 70:
```suggestion
> Note: Calls to Seaport that would fulfill or match a collection of advanced orders can be monitored and where there are unused offer items, it's possible for a third party to claim them. Anyone can monitor the mempool to find calls to `matchOrders` or `matchAdvancedOrders` without "ad-hoc" orders (where the offerer is the caller, hence does not require a signature) and calculate if there are any unused offer item amounts.  If there are unused offer item amounts, the third party can frontrun the transaction and supply themselves as the recipient, thereby allowing that third party to claim the unused offer items for themselves. A Seaport app or a zone could prevent this, or the fulfiller can utilize a private mempool, but by default it's possible.
```

PR #970, opened 2023-02-10
Fix: Trigger Accumulators Before Post Execution Hooks are called

androolloyd:
## Motivation
When creating an order, you can supply a conduitKey, this key represents which Conduit the Consideration can communicate with to transfer the offerer's assets.

When the conduitKey is bytes32(0) seaport is the acting Conduit.

When this is the case, the transactions to settle the order are executed as they are processed inside the execution array loop atomically, when a conduitKey is set an accumulator(an array of transactions to execute) is armed and later triggered.

As of Seaport 1.2, there is a new post execution hook for restricted orders that gets triggered inside the order validity array loop.

When Seaport is the conduit, the order is atomically settled at the time of the validateOrder call, and if a conduitKey is set, then the order is not settled at the point of execution but after all the validateOrder calls have been made and the accumulator has been triggered.

## Solution

Trigger the accumulator first between the execution loop and order validity loop in addition to after

### DIFF
--- a/contracts/lib/OrderCombiner.sol
+++ b/contracts/lib/OrderCombiner.sol
@@ -746,6 +746,9 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
             }
         }
 
+        // Trigger execution prior to post-execution checks.
+        _triggerIfArmed(accumulator);
+
         // Skip overflow checks as all for loops are indexed starting at zero.
         unchecked {
             // duplicate recipient address to stack to avoid stack-too-deep


contracts/lib/OrderCombiner.sol
```diff
@@ -727,6 +727,9 @@ contract OrderCombiner is OrderFulfiller, FulfillmentApplier {
             }
         }
 
+        //trigger the execution of the orders first so we have atomic settlement of the orders
```
0age, line 730: Let’s revise this comment to `// Trigger execution prior to post-execution checks.`

0age:
This was a great find! We ended up just performing an additional loop through each order at the end if any restricted or contract orders are present, as that way you can rely on _every_ transfer having occurred: https://github.com/ProjectOpenSea/seaport/blob/85f395a3a49ea1b48d945ae975ef16127320ee76/contracts/lib/OrderCombiner.sol#L882

PR #978, opened 2023-02-14
Transfer Validation Zone/Offerer

BCLeFevre:
## Motivation

Zone framework for testing an issue with order fulfillment and the timing when our accumulator is triggered in v1.2


DJViau (approved):
Opened a PR with a tiny bit of cleanup [here](https://github.com/ProjectOpenSea/seaport/pull/980).

Great work, thanks for knocking this out in such short order.

PR #979, opened 2023-02-14
use big switch instead of typehash directory

0age:


contracts/lib/ConsiderationBase.sol
```diff
@@ -277,14 +294,185 @@ contract ConsiderationBase is
...
+                            ),
+                            mul(
+                                eq(treeHeight, 18),
+                                BulkOrder_Typehash_Height_Eighteen
+                            )
+                        )
+                    }
+                }
+                default {
+                    switch lt(treeHeight, 21)
```
0age, line 437:
```suggestion
                    switch lt(treeHeight, 22)
```

PR #988, opened 2023-02-14
include metadata hash again

0age:


0age:
eh this'll actually make reproducing create2 deploys harder

PR #1015, opened 2023-03-08
Seaport v1.5

0age:
This PR fixes a minor bug, introduced in Seaport v1.3, that causes the `amount` parameter on spent offer item amounts reported to zones and contract offerers to be overwritten by the `recipient` address on spent offer items with excess (or unspent) item amounts when using `match` or `fulfillAvailable` methods.

As part of these combined fulfillment methods, Seaport iterates over each supplied order and asserts that all consideration item amounts have been fully met while also checking for unspent offer item amounts. If an unspent offer item amount is located, it typecasts the offer item in-place to a received item and writes the designated recipient to memory, overwriting the `endAmount` parameter on the offer item, and uses that received item to perform a transfer, sending the excess offer item amounts to the designated recipient.

Once all transfers have been fully processed, Seaport then iterates over each supplied order another time and performs a call to `validateOrder` on the zone (for restricted orders) or `ratifyOrder` on the offerer (for contract orders), encoding the call based on the order data in memory. Specifically, the `endAmount` parameter is used to encode the amount for each item (the full amount is originally written to both `startAmount` _and_ `endAmount`, but `startAmount` is used to debit offer items and credit consideration items during fulfillment). As offer items with excess item amounts have had this value overwritten with the `recipient` value, Seaport erroneously reports the `amount` on these offer items to the zone or contract offerer as being equal to the recipient address.

The fix proposed in this PR reads the `endAmount` value and places the result on the stack prior to overwriting it with the `recipient` value, then restores the original `endAmount` value immediately after performing the associated transfer.

Aside from this patch, the only change to the contracts in this PR is to bump the version to 1.5.


0age:
Deployed: https://etherscan.io/address/0x00000000000000adc04c56bf30ac9d3c0aaf14dc#code

d1ll0n (approved):
🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥

PR #1016, opened 2023-03-08
Dan/2023/03/fuzz testing effort on complex functions

DJViau:
## Motivation

Better test coverage of complex interactions.

## Solution

Fuzz testing.


test/foundry/zone/TestTransferValidationZoneOfferer.t.sol
```diff
@@ -2563,6 +2654,168 @@ contract TestTransferValidationZoneOffererTest is BaseOrderTest {
...
+                    );
+
+                    // Build first restricted order components, remove conduit key.
+                    infra.orderComponents = OrderComponentsLib
+                    .fromDefault(VALIDATION_ZONE)
+                    .withOffer(infra.offerArray)
+                    .withConsideration(infra.considerationArray)
+                    // TODO: REMOVE: Come back and figure out how to fuzz the
+                    //               conduit key.
+                        .withConduitKey(bytes32(0))
```
jameswenzel, line 2711: test infra only has one conduit set up; in other tests i believe we fuzz on "conduit: yes/no" and conditionally use `bytes32(0)` or `conduitKey[One]` (the "new" test base classes just call it `conduitKey`)

DJViau: Thanks. I was going to implement that pattern (as [here](https://github.com/ProjectOpenSea/seaport/pull/1016/files#diff-945414f958c6be186ee73df95b1ea91413b7fc93e9464c54f2565917ffd4f5deR1413)), but I also kind of wanted to play around to see if I could get a test that uses multiple conduits running

PR #1023, opened 2023-03-13
Test annotations + explicit named args

horsefacts:
## Motivation
I found it a little difficult to get my bearings in the tests, and jumped around a lot looking up helper functions and structs. I started these annotations for myself, but maybe this pattern will be useful everywhere.

## Solution

- Add top-level comments to tests with a short summary of each test case. This includes a short summary of what it tests, plus a brief description of its "arrange," "act," "assert" steps.
- Use named args syntax for structs and calls to helper functions. This makes it easier to follow what each arg represents without having to jump outside the test context. 
- Add internal comments to tests where helpful.
- Factor out repeated references (like `startTime` in the ascending/descending offer/consideration tests)


horsefacts:
After spending some more time around the codebase, I think the new builder libraries mostly address the test readability concerns I had here. Probably a good idea to make sure new tests follow these guidelines, but I don't think the best place to focus is updating the old ones right now. Closing this one for now, will re-open later if it makes sense.

PR #1025, opened 2023-03-14
Move asc/desc fulfillOrder tests

horsefacts:
Convert two simple `fulfillOrder` tests (ascending/descending offers and considerations) to the new builder style.


0age (approved):

horsefacts:
Still want to confirm that the events here are being caught correctly. Will do that and rebase when I get a sec.

PR #1031, opened 2023-03-16
MOAT execution engine

horsefacts:
**MOAT execution engine — first pass**

- `FuzzHelpers`: Stateless helper library to detect structure/state/type/family/quantity for a given order configuration.
- `FuzzEngineLib`: Stateless helpers for the `FuzzEngine`. Includes functions to select a Seaport function to call for the given order configuration.
- `FuzzEngine`: A base contract for MOAT tests. Provides two functions: 
  - `exec(TestContext)`: takes a `TestContext` struct including a list of orders, a Seaport interface, fuzz parameters, and an array of selectors of post-execution "checks." Selects a Seaport method and calls it for the given order configuration, including any extra data passed in the test context or individual order context.
  - `checkAll(TestContext)`: performs all registered "checks." Checks should be public functions that accept a `TestContext` as their only argument.
  
  See the inline comments for more detailed documentation.


test/foundry/new/helpers/MOATHelpers.sol
```diff
@@ -0,0 +1,276 @@
...
+        bool isContractOrder = order.order.parameters.orderType ==
+            OrderType.CONTRACT;
+
+        // If any non-contract item has criteria, it's advanced,
+        if (hasCriteria) {
+            // Unless it's a contract order
+            if (isContractOrder) {
+                // And the contract order's critera are all zero
+                if (hasNonzeroCriteria) {
+                    return Structure.ADVANCED;
```
0age, line 202: STANDARD

test/foundry/new/helpers/MOATHelpers.sol
```diff
@@ -0,0 +1,276 @@
...
+        MOATOrder memory order
+    ) internal pure returns (bool hasCriteria, bool hasNonzeroCriteria) {
+        // Check if any offer item has criteria
+        OfferItem[] memory offer = order.order.parameters.offer;
+        for (uint256 i; i < offer.length; ++i) {
+            OfferItem memory offerItem = offer[i];
+            ItemType itemType = offerItem.itemType;
+            hasCriteria = (itemType == ItemType.ERC721_WITH_CRITERIA ||
+                itemType == ItemType.ERC1155_WITH_CRITERIA);
+            if (offerItem.identifierOrCriteria != 0) hasNonzeroCriteria = true;
```
0age, line 225: can return earlt

test/foundry/new/helpers/MOATHelpers.sol
```diff
@@ -0,0 +1,276 @@
...
+    function toMOATOrder(
+        AdvancedOrder memory order
+    ) internal pure returns (MOATOrder memory) {
+        return
+            MOATOrder({
+                order: order,
+                context: MOATOrderContext({
+                    signature: bytes(""),
+                    counter: 0,
+                    fulfillerConduitKey: bytes32(0),
```
0age, line 257: these should be in a high-level context struct

test/foundry/new/helpers/MOATHelpers.sol
```diff
@@ -0,0 +1,276 @@
...
+     * @param order an AdvancedOrder struct.
+     */
+    function toMOATOrder(
+        AdvancedOrder memory order
+    ) internal pure returns (MOATOrder memory) {
+        return
+            MOATOrder({
+                order: order,
+                context: MOATOrderContext({
+                    signature: bytes(""),
```
0age, line 255: this can just be on the order

test/foundry/new/helpers/MOATHelpers.sol
```diff
@@ -0,0 +1,276 @@
...
+    /**
+     * @dev Convert an AdvancedOrder to a MOATOrder, filling in the context
+     *      with empty defaults.
+     * @param order an AdvancedOrder struct.
+     */
+    function toMOATOrder(
+        AdvancedOrder memory order
+    ) internal pure returns (MOATOrder memory) {
+        return
+            MOATOrder({
```
0age, line 252: should also include desired results

test/foundry/new/helpers/MOATHelpers.sol
horsefacts: Can you take another look at this when you have a chance? I think the logic here might actually be correct, there's just a few layers of negatives that make it difficult to read.

From the doc:
- Order cannot contain any items with itemType ERC721WithCriteria or ERC1155WithCriteria, *unless*:
    - The order is a contract order and
    - All criteria-based items have `identifierOrCriteria == 0`
    - The contract offerer can successfully generate an order that resolves every criteria-based item without any context / extraData

```solidity
        // Order cannot contain any items with itemType ERC721WithCriteria or ERC1155WithCriteria, *unless*
        if (hasCriteria) {
            // The order is a contract order
            if (isContractOrder) {
                // and all criteria-based items have `identifierOrCriteria == 0`
                if (hasNonzeroCriteria) {
                    return Structure.ADVANCED;
                }
            } else {
                return Structure.ADVANCED;
            }
        }
```

Because these are basically guards, the default case here is `Structure.STANDARD`. If the logic is right, I'll try and detangle it—checking if there are nonzero criteria in an "unless" case is a confusing way to do this.

0age: ah that makes sense!

test/foundry/new/helpers/FuzzEngine.sol
```diff
@@ -0,0 +1,257 @@
...
+}
+
+/**
+ * @notice Base test contract for FuzzEngine. Fuzz tests should inherit this.
+ *         Includes the setup and helper functions from BaseOrderTest.
+ *
+ *         Engine lifecycle:
+ *           - generate a TestContext. This struct includes:
+ *           - exec(context)
+ *           - checkAll(context)
```
horsefacts, line 94: Remove for now.

PR #1048, opened 2023-03-21
Dan/2023/03/add fulfill basic and match to moat

DJViau:
## Motivation

We want it all.

## Solution

Add the remaining.

### DIFF
--- a/contracts/helpers/sol/lib/AdvancedOrderLib.sol
+++ b/contracts/helpers/sol/lib/AdvancedOrderLib.sol
@@ -2,13 +2,17 @@
 pragma solidity ^0.8.17;
 
 import {
+    AdditionalRecipient,
     AdvancedOrder,
+    BasicOrderParameters,
     ConsiderationItem,
     OfferItem,
     Order,
     OrderParameters,
     OrderType
 } from "../../../lib/ConsiderationStructs.sol";
+
+import { BasicOrderType } from "../../../lib/ConsiderationEnums.sol";
 
 import { OrderParametersLib } from "./OrderParametersLib.sol";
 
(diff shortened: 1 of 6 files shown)


contracts/helpers/sol/lib/OrderLib.sol
```diff
@@ -284,4 +288,120 @@ library OrderLib {
...
+     *
+     * @return advancedOrders the AdvancedOrders
+     */
+    function toAdvancedOrders(
+        Order[] memory orders,
+        uint120 numerator,
+        uint120 denominator,
+        bytes memory extraData
+    ) internal pure returns (AdvancedOrder[] memory advancedOrders) {
+        for (uint256 i = 0; i < orders.length; i++) {
```
jameswenzel, line 308: You'll need to allocate `advancedOrders` first

contracts/helpers/sol/lib/OrderLib.sol
```diff
@@ -284,4 +288,120 @@ library OrderLib {
...
+     * @dev Converts an Order to an AdvancedOrders array.
+     *
+     * @param order the Order to convert
+     * @param numerator the numerator to set for all
+     * @param denominator the denominator to set for all
+     * @param extraData the extra data to set for all
+     *
+     * @return advancedOrders the AdvancedOrders
+     */
+    function toAdvancedOrders(
```
jameswenzel, line 328: personally more in favor of `SeaportArrays.AdvancedOrders(order.toAdvancedOrder())` than a dedicated lib fn for this – I still want to write a script to generate these libs and it'll need special logic for each non-generated struct-specific helper fn

DJViau: roger, will update

contracts/helpers/sol/lib/AdvancedOrderLib.sol
```diff
@@ -365,4 +369,80 @@ library AdvancedOrderLib {
...
+
+    /**
+     * @dev Converts an AdvancedOrder to a BasicOrderParameters.
+     *
+     * @param advancedOrder  the AdvancedOrder to convert
+     * @param basicOrderType the BasicOrderType to convert to
+     *
+     * @return basicOrderParameters the BasicOrderParameters
+     */
+    function toBasicOrderParameters(
```
jameswenzel, line 381: we may want to consider adding revert logic if an order is unable to be converted, but I'm fine without it for now

PR #1064, opened 2023-03-24
Dan/2023/03/add a check registration step

DJViau:
Same as [this PR](https://github.com/ProjectOpenSea/seaport/pull/1063/files) (closed in favor of this one), but it also adds a check registration step.

### DIFF
--- a/test/foundry/new/FuzzEngine.t.sol
+++ b/test/foundry/new/FuzzEngine.t.sol
@@ -6,6 +6,7 @@ import "seaport-sol/SeaportSol.sol";
 import "forge-std/console.sol";
 
 import {
+    CheckHelpers,
     FuzzTestContext,
     FuzzParams,
     FuzzEngine,
@@ -17,7 +18,6 @@ import {
     HashValidationZoneOfferer
 } from "../../../contracts/test/HashValidationZoneOfferer.sol";
 import { AdvancedOrder, FuzzHelpers } from "./helpers/FuzzHelpers.sol";
-import { CheckHelpers } from "./helpers/FuzzSetup.sol";
 
 contract FuzzEngineTest is FuzzEngine {
     using OfferItemLib for OfferItem;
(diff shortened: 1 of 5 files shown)


test/foundry/new/FuzzEngine.t.sol
```diff
@@ -17,6 +17,7 @@ import {
     HashValidationZoneOfferer
 } from "../../../contracts/test/HashValidationZoneOfferer.sol";
 import { AdvancedOrder, FuzzHelpers } from "./helpers/FuzzHelpers.sol";
+import { CheckHelpers } from "./helpers/FuzzSetup.sol";
```
horsefacts, line 20: Might make sense to move this to `FuzzEngineLib`, or `FuzzTestContextLib`, what do you think?

test/foundry/new/FuzzEngine.t.sol
```diff
@@ -493,6 +492,9 @@ contract FuzzEngineTest is FuzzEngine {
                 orders[0].toBasicOrderParameters(orders[0].getBasicOrderType())
             );
 
+        context.registerCheck(this.check_orderFulfilled.selector);
+        context.registerCheck(this.check_orderStatusFullyFilled.selector);
```
horsefacts, line 496: `registerCheck` returns the `TestContext`, so these are chainable if you want, e.g.:

```
context.registerCheck(this.check_orderFulfilled.selector).registerCheck(this.check_orderStatusFullyFilled.selector);
```

horsefacts (approved):

test/foundry/new/FuzzEngine.t.sol
DJViau: I forgot that.  I just changed it and prettier makes it uglier.  imo it's more readable as-is, but I'm happy to chain if we want to adopt chain-when-possible or something like that as a convention.

test/foundry/new/FuzzEngine.t.sol
DJViau: Yeah, sg.  FuzzEngineLib seems like he right play based on gut feel.  Happy to reshuffle as necessary.  Will update and then merge into your branch

PR #1104, opened 2023-04-04
update inscribers and add unavailable cases

DJViau:


test/foundry/new/helpers/FuzzGenerators.sol
```diff
@@ -327,36 +331,47 @@ library AdvancedOrdersSpaceGenerator {
...
         } else if (reason == UnavailableReason.STARTS_IN_FUTURE) {
-            order = order.withGeneratedTime(
+            parameters = parameters.withGeneratedTime(
                 Time.STARTS_IN_FUTURE,
                 context
             );
+        } else if (reason == UnavailableReason.CANCELLED) {
+            order.inscribeOrderStatusCanceled(true, context.seaport);
+        } else if (reason == UnavailableReason.ALREADY_FULFILLED) {
+            order.inscribeOrderStatusNumeratorAndDenominator(
```
0age, line 364: Shouldn’t this inscribe 1 + 1? We want to update the state to reflect that it’s fully filled, not that it’s already filled with the numerator / denominator we’re attempting to fill?

PR #1108, opened 2023-04-06
Fuzz ascending/descending offers

horsefacts:
- Fuzz ascending/descending amounts for offer items
- Extract and share `_locateCurrentAmount` to `FuzzHelpers`
- Warp to a realistic timestamp (Jan 1 2023) before tests.
- Limit fuzzed timestamps to `type(uint40).max`

### DIFF
--- a/test/foundry/new/helpers/FuzzEngineLib.sol
+++ b/test/foundry/new/helpers/FuzzEngineLib.sol
@@ -3,7 +3,12 @@ pragma solidity ^0.8.17;
 
 import "seaport-sol/SeaportSol.sol";
 
-import { Family, FuzzHelpers, Structure } from "./FuzzHelpers.sol";
+import {
+    Family,
+    FuzzHelpers,
+    Structure,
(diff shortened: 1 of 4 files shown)


test/foundry/new/helpers/FuzzEngineLib.sol
```diff
@@ -233,38 +240,44 @@ library FuzzEngineLib {
...
-
                 if (item.itemType == ItemType.NATIVE) {
-                    value += item.startAmount;
+                    if (item.startAmount != item.endAmount) {
+                        value += _locateCurrentAmount(
+                            item.startAmount,
+                            item.endAmount,
+                            orderParams.startTime,
+                            orderParams.endTime,
+                            true
```
0age, line 258: I think this should be false for offer items?

horsefacts: Ah, whoops, looks like they're reversed.

PR #1109, opened 2023-04-06
Fuzz ascending/descending considerations

horsefacts:
- Fuzz ascending/descending consideration item amounts
- Add `isAvailable` helper to `OrderParametersLib` 
- Skip calculating spent/received items for unavailable orders.

### DIFF
--- a/contracts/helpers/sol/lib/OrderParametersLib.sol
+++ b/contracts/helpers/sol/lib/OrderParametersLib.sol
@@ -500,6 +500,6 @@ library OrderParametersLib {
     ) internal view returns (bool) {
         return
             block.timestamp >= parameters.startTime &&
-            block.timestamp <= parameters.endTime;
+            block.timestamp < parameters.endTime;
     }
 }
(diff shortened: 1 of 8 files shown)


contracts/helpers/sol/lib/OrderParametersLib.sol
```diff
@@ -494,4 +494,12 @@ library OrderParametersLib {
         components.conduitKey = parameters.conduitKey;
         components.counter = counter;
     }
+
+    function isAvailable(
+        OrderParameters memory parameters
+    ) internal view returns (bool) {
+        return
+            block.timestamp >= parameters.startTime &&
+            block.timestamp <= parameters.endTime;
```
0age, line 503:
```suggestion
            block.timestamp < parameters.endTime;
```

horsefacts: Should whether the order is cancelled or not also be part of this definition? Don't think we need it right now, but we'd have to make a call to get it.

horsefacts: In fact, maybe we should use the definition from [ZoneParametersLib](https://github.com/ProjectOpenSea/seaport/blob/7fd32462511a82b0b13860d817efc3ba68364a4d/contracts/helpers/sol/lib/ZoneParametersLib.sol#L221) everywhere, which also checks `totalFilled`.

PR #1124, opened 2023-04-10
Fuzz on top level caller

horsefacts:
🚧


test/foundry/new/helpers/FuzzGenerators.sol
```diff
@@ -2031,6 +2041,31 @@ library FulfillmentRecipientGenerator {
...
 }
 
+library CallerGenerator {
+    function generate(
+        Caller caller,
+        FuzzGeneratorContext memory context
+    ) internal view returns (address) {
+        if (caller == Caller.TEST_CONTRACT) {
+            return address(this);
+        } else if (caller == Caller.OFFERER) {
```
0age, line 2051: Not sure this one makes sense as there may be multiple offerers on distinct orders; we might want to remove this from the top level of the test context entirely

PR #1135, opened 2023-04-11
scaffold failure tests

0age:


test/foundry/new/helpers/FuzzMutationSelectorLib.sol
```diff
@@ -0,0 +1,151 @@
...
+    // BadSignatureV, // EOA signature has bad v value
+    // BadContractSignature_BadSignature, // 1271 call to offerer, signature tampered with
+    // BadContractSignature_ModifiedOrder, // Order with offerer with code tampered with
+    // BadContractSignature_MissingMagic, // 1271 call to offerer, no magic value returned
+    // ConsiderationLengthNotEqualToTotalOriginal, // Tips on contract order or validate
+    // BadFraction_PartialContractOrder, // Contract order w/ numerator & denominator != 1
+    // BadFraction_NoFill, // Order where numerator = 0
+    // BadFraction_Overfill, // Order where numerator > denominator
+    length // NOT A FAILURE; used to get the number of failures in the enum
+}
```
DJViau, line 29: I'm still reading through this for the first time, so not sure it's even relevant, just wanted to make sure that `ErrorSpaceEnums.sol` was on your radar

test/foundry/new/helpers/FuzzMutations.sol
```diff
@@ -5,17 +5,126 @@ import { Test } from "forge-std/Test.sol";
...
+            context.orderHashes[orderIndex]
+        );
+
+        if (isValidated) {
+            return false;
+        }
+
+        return true;
+    }
+}
```
DJViau, line 52: Aren't these booleans flipped?  Or does something down stream make it make sense?

Seems like the name `ineligibleForInvalidSignature` should return `true` in cases where we're NOT going to be able to trigger a failure by passing in an invalid sig.

PR #1137, opened 2023-04-12
restructure the test context struct

DJViau:
Buy some headroom


horsefacts:
Not a huge fan of the name `executionState` but I don't have a better suggestion!

PR #1149, opened 2023-04-14
Initial zone and contract offerer failures

jameswenzel:


0age:
This one's mostly already in but has fallen kind of far out of sync with what's already merged in given all the changes related to scoping by order hash and such, it's really just the change to provide a full ReceivedItem instead of an amount that's not currently included; let's open a new PR with that change when the time comes

PR #1164, opened 2023-04-19
InvalidMsgValue error cases

horsefacts:
Add failure tests for `InvalidMsgValue`.


test/foundry/new/helpers/FuzzMutations.sol
```diff
@@ -142,6 +145,23 @@ library MutationFilters {
         return false;
     }
 
+    function ineligibleForInvalidMsgValue(
```
horsefacts, line 148: This should probably be a generic `ineligibleWhenNotBasic`

PR #1179, opened 2023-04-21
Unmet consideration item

horsefacts:
Error cases for `ConsiderationItemNotMet`. Note: we are probably over-filtering by requiring numerator = denominator, but I couldn't work out how to prevent `InexactFraction()` errors.

### DIFF
--- a/test/foundry/new/helpers/FuzzMutations.sol
+++ b/test/foundry/new/helpers/FuzzMutations.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.17;
-
+import "forge-std/console.sol";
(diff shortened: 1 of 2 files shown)


test/foundry/new/helpers/FuzzMutations.sol
```diff
@@ -1259,6 +1260,45 @@ library MutationFilters {
...
+            return true;
+        }
+
+        // TODO: Probably overfiltering
+        if (order.numerator != order.denominator) {
+            return true;
+        }
+
+        // Must not be a contract order
+        if (order.parameters.orderType == OrderType.CONTRACT) {
```
0age, line 1286: Not sure why this is a requirement, but we'll leave it for the time being

PR #1201, opened 2023-04-29
Update README.md with support for Moonbeam and Moonriver

x676f64:
Adding Moonbeam and Moonriver to the `Deployments by EVM Chain` section as I have completed deployment and contract verification for Seaport v1.5.

## Motivation

https://twitter.com/z0age/status/1651960136389525507


0age:
Mind removing the testnet deploy from the list? Maybe we could add a separate table in the docs/ folder for tracking those?

PR #1221, opened 2023-05-05
update forge-std, remove now-included makeAccount

ryanio:


ryanio:
would love some help figuring out what's going wrong with the 1271 forge tests, since I'm importing seaport into seadrop I'm getting errors with `makeAccount` now being available in latest forge-std lib, i can confirm this PR fixes the downstream issue I'm facing

PR #1238, opened 2023-05-12
SeaportOrderHelper contract

horsefacts:
Add `SeaportOrderHelper`. This read-only helper contract generates additional information useful for fulfilling Seaport orders. Given an array of `AdvancedOrder` structs and external parameters like caller/recipient/native tokens supplied, `SeaportOrderHelper` will:

- Validate the orders and return associated errors and warnings.
- Recommend a fulfillment method (i.e. "call this Seaport function").
- Generate and suggest fulfillments.
- Calculate and return `Execution` and `OrderDetails` structs for the given orders.
- Optionally generate criteria resolvers based on provided constraints.

`CriteriaConstraint` is a new struct similar to `CriteriaResolver` that defines criteria in terms of token IDs rather than Merkle roots/proofs. A `CriteriaConstraint` specifies:

- An order index, side, and item index describing which order item is associated with the constraint.
- An array of eligible token IDs to use to generate criteria.
- The actual token ID that will be provided for the given item at fulfillment time.

The order helper will calculate criteria Merkle roots/proofs for each constraint, modify orders in place to add the roots as each item's `identifierOrCriteria`, and return the calculated proofs/criteria resolvers for use in fulfillment. 

Other stuff:
- This bump the `seaport-sol` version to `^1.5.3` (this release includes updated NPM compatible imports) and updates all `seaport-sol` imports to NPM-style. (Unfortunately, this means a lot of import changes in the diff).
- This PR includes a pure `MerkleLib` library adapted from Murky, using higher order functions rather than an abstract contract + virtual functions to specify the hash function. It might be nice to upstream this if there's interest.


contracts/helpers/order-helper/lib/OrderStructureLib.sol
```diff
@@ -0,0 +1,516 @@
...
+
+        // Order cannot be partially filled.
+        ConsiderationInterface seaportInterface = ConsiderationInterface(
+            seaport
+        );
+        uint256 counter = seaportInterface.getCounter(order.parameters.offerer);
+        OrderComponents memory orderComponents = order
+            .parameters
+            .toOrderComponents(counter);
+        bytes32 orderHash = seaportInterface.getOrderHash(orderComponents);
```
0age, line 271: shouldn't this be operating on the "tip-neutralized" order hash?

contracts/helpers/order-helper/lib/OrderAvailabilityLib.sol
```diff
@@ -0,0 +1,80 @@
...
+    ) internal view returns (bool) {
+        return unavailableReason(order, seaport) == UnavailableReason.AVAILABLE;
+    }
+
+    /**
+     * @notice Returns the order's UnavailableReason. Available orders will
+     *         return UnavailableReason.AVAILABLE to indicate that they are
+     *         available for fulfillment.
+     */
+    function unavailableReason(
```
0age, line 35: don't we need to also account for an order that's unavailable due to a reverting generateOrder check? Maybe there's not a straightforward way to check for that?

contracts/helpers/order-helper/lib/OrderStructureLib.sol
horsefacts: Interesting. Should this behave differently than the [fuzz engine](https://github.com/ProjectOpenSea/seaport/blob/cbcd738ee7be5c4178d550b843b71c0bb18d2d04/test/foundry/new/helpers/FuzzHelpers.sol#L510)? It's also using plain order hash rather than the tip neutralized version.

0age: might honestly be a "bug" in the fuzz engine (causing us to not pick basic orders as often as we could)?

contracts/helpers/order-helper/lib/OrderAvailabilityLib.sol
horsefacts: At the moment, this doesn't support contract orders at all—it just reverts early if any of the provided orders are contract orders.

PR #1249, opened 2023-05-31
update oz lib to latest

ryanio:
Having some [issues](https://github.com/ProjectOpenSea/seadrop/commit/b286ca1e303b6adfcbd6ad3a6cb0bcb01fe4b8c1) with foundry remappings, so would be great to get openzepplin-contracts lib updated to latest so can use it within seadrop directly without worrying about conflicting contract versions


ryanio:
nvm

PR #1293, opened 2023-11-17
discuss: potential direction for refactoring differential tests

anna-carroll:
## Motivation

In conversations with @jameswenzel , he said that Foundry had fixed the issue discussed here https://github.com/ProjectOpenSea/seaport/discussions/809 - that rolling back chain state using vm.snapshot no longer erases failed test assertions. 

If this is true, the `stateless` modifier is no longer necessary, and it is possible to refactor Differential tests to use a simpler setup. 

## Solution

Here I've put forth a potential direction for refactoring Differential tests. I can extend to other tests currently using `stateless` if we align on the approach.


0age:
This does seem much cleaner to me!

anna-carroll:
> This does seem much cleaner to me!

Update on my side: not sure `vm.revertTo` is very stable https://github.com/foundry-rs/foundry/issues/6355

Evalir:
Hey hey, just an update here @anna-carroll / @0age : we're aware of the issues with `snapshot`/`revertTo`—@mattsse tracked this particular bug down and the fix is on https://github.com/foundry-rs/foundry/pull/6451 — let us know if this is sufficient to unblock this, and if not let us know what else should we fix. We don't want foundry being a blocker for this!

anna-carroll:
> Hey hey, just an update here @anna-carroll / @0age : we're aware of the issues with `snapshot`/`revertTo`—@mattsse tracked this particular bug down and the fix is on [foundry-rs/foundry#6451](https://github.com/foundry-rs/foundry/pull/6451) — let us know if this is sufficient to unblock this, and if not let us know what else should we fix 🙏

This is amazing @Evalir I think I've been able to verify that the fix works reliably! 

I've started in on deprecating the stateless modifier in favor of using snapshots.

PR #1295, opened 2023-12-24
fix: run linter

anna-carroll:
## Motivation

It seems the code in `main` has some lint warnings when running `yarn lint:check.`

## Solution

Run `yarn lint:fix`


ryanio:
thanks for the PR! we'll probably do lint:fixes during Seaport 1.6 development (currently on Seaport 1.5) so that the 1.5 code matches the deployed version.

ryanio:
actually i'll reopen this to fix other PRs, I'm not sure why the linting would have changed so much, maybe something related to line length. @0age anything in particular you want to do here?

